PR #11, opened 2021-08-10
docs: change to mkdocs and google style

hfudev:
also update docs


igrr:
Perhaps now that we aren't using the Gitlab repo, we should also add an action to deploy the docs somewhere public? Like https://github.com/mhausenblas/mkdocs-deploy-gh-pages.

hfudev:
@igrr since we're using https://docs.espressif.com/projects/pytest-embedded/en/latest/ to host it and `readthedocs` also supports `mkdocs`, should we try to deploy to the same place?

igrr:
My bad, i forgot that we host this on RTD already!

RTD has "build type" as a global setting only, so I have switched it from Sphinx to mkdocs. This means that the docs on the master branch are likely broken now. The docs for this branch on RTD currently fails because of the missing `pytest-embedded` dependency. Once it is fixed, the docs will be available at https://readthedocs.com/projects/espressif-pytest-embedded/builds/695925/.

hfudev:
@igrr here I'm using the `.readthedocs.yaml` for the build process and it's working. PTAL. https://docs.espressif.com/projects/pytest-embedded/en/docs-update_docs_and_examples/

PR #13, opened 2021-09-15
ci: enable custom runners for serial tests

hfudev:


hfudev:
@tomassebestik @igrr PTAL. For now the cache for rpi built python is working.

hfudev:
@igrr I add the `gen_esp32part.py` back to let people run the tests without IDF.

PR #16, opened 2021-11-16
feat: return `re.Match` if `dut.expect()` succeeded

hfudev:


DCSBL:
Is there an example how this can be used?

hfudev:
@DCSBL Hi! I've improved the example here to explain from the script aspect https://github.com/espressif/pytest-embedded/commit/079406f3946b8fa78dfaadde4cddcc8c4af446de

for a specific use case, you can use this value as a kind of benchmark and save this to the junit report by pytest fixture https://docs.pytest.org/en/6.2.x/reference.html#record-property

DCSBL:
Thanks, that did it. I use this to run Unity Unit test and I now can detect if the tests have been passed of failed. I've used this if anyone is interested: 

```python
import pytest
from pytest_embedded.log import live_print_call

import logging

def test_unit_test_results(dut):
    res = dut.expect(r'([0-9]*) Tests ([0-9]*) Failures ([0-9]*) Ignored') # Searches for something like '<num> Tests <num> Failures <num> Ignored'
    
    tests_num = res.group(1).decode("utf8")
    failed_num = res.group(2).decode("utf8")
    ignored_num = res.group(3).decode("utf8")
    
    logging.info(f'Total tests: {tests_num}')
    logging.info(f'Failed tests: {failed_num}')
    logging.info(f'Ignored tests: {ignored_num}')
    
    if (int(failed_num) != 0):
        pytest.fail()    
    
    dut.expect('OK')
```

Resulting into:
```bash
| 2021-11-18 09:05:56 INFO [qemu] 8 Tests 1 Failures 0 Ignored
| -------------------------------- live log call ---------------------------------
| 2021-11-18 09:05:56 INFO Total tests: 8
| 2021-11-18 09:05:56 INFO Failed tests: 1
| 2021-11-18 09:05:56 INFO Ignored tests: 0
| FAILED                                                                   [100%]
```

igrr:
Thanks @DCSBL, that's useful. Since Unity is very common, I think we'll add built-in functionality to pytest-embedded to parse unity reports (including log and error messages).

PR #22, opened 2021-11-25
feat: add `expect_list` and `expect_exact`

hfudev:
and add related tests

### DIFF
--- a/pytest-embedded/pytest_embedded/dut.py
+++ b/pytest-embedded/pytest_embedded/dut.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Match, Optional
+from typing import AnyStr, Match, Optional, Union
 
 import pexpect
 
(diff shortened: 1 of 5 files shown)


pytest-embedded/pytest_embedded/dut.py
```diff
@@ -33,15 +33,70 @@ def write(self, *args, **kwargs) -> None:
...
+        Expect from `pexpect_proc`. All the arguments would pass to `pexpect.expect_list()`.
+
+        Returns:
+            AnyStr: if you're matching pexpect.EOF or pexpect.TIMEOUT to get all the current buffers.
+
+        Returns:
+            re.Match: if matched given compiled regex.
+
+        Notes:
+            For the first argument, which is the regex list, you should pass the compiled regex with bytes.
```
igrr, line 89: It looks a bit odd that this is the only case where we need to pass bytes rather than string as an argument?

pytest-embedded/pytest_embedded/dut.py
```diff
@@ -33,15 +33,70 @@ def write(self, *args, **kwargs) -> None:
...
+                return self.pexpect_proc.before.rstrip()
+
+            return self.pexpect_proc.match
+
+    def expect_list(self, *args, **kwargs) -> Union[Match, AnyStr]:
+        """
+        Expect from `pexpect_proc`. All the arguments would pass to `pexpect.expect_list()`.
+
+        Returns:
+            AnyStr: if you're matching pexpect.EOF or pexpect.TIMEOUT to get all the current buffers.
```
igrr, line 83: The test case seems to indicate that the result is actually `bytes`?

pytest-embedded/pytest_embedded/dut.py
hfudev: yeah... cause the [`expect_list`](https://pexpect.readthedocs.io/en/stable/api/pexpect.html#pexpect.spawn.expect_list) would take compiled regex as input, and we're using bytes in our buffer.

If we only pass compiled string regex, it would raise exception `TypeError: cannot use a string pattern on a bytes-like object`.

pytest-embedded/pytest_embedded/dut.py
hfudev: according to https://docs.python.org/3/library/typing.html#typing.AnyStr, `AnyStr` refers to `bytes` or `str`

PR #23, opened 2021-11-26
feat: real-time logging

hfudev:
Before the stdout output is always lagging than real time.

this works but would cause the logging system ugly, with no indentation as the same length as the logging format string

TODO:
- [x] use logging adapter to improve this


hfudev:
ah, this feature is provided by `pytest` already: https://docs.pytest.org/en/6.2.x/reference.html#confval-log_auto_indent

PR #24, opened 2021-12-07
feat(esp): simplify esptool call process

hfudev:


pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -107,9 +95,10 @@ def __init__(self, attributes):
                 flash_files.append((address, open(nvs_file.name, 'rb')))
 
         try:
-            self.stub.change_baud(baud_rate)
-            esptool.detect_flash_size(self.stub, flash_args)
-            esptool.write_flash(self.stub, flash_args)
+            self.esp.change_baud(921600)  # higher baudrate for data transferring
```
igrr, line 98: Is it possible to get the baud rate value from ESPBAUD environment variable, if it is set? (Or add some option to set higher baud rate). In IDF, ESPPORT and ESPBAUD variables can be used to give a hint to the flasher about the port and baud rate to use. I usually set ESPBAUD=2000000 in my profile script since the majority of boards work on my computer at this baud rate.

hfudev: make sense. I'll follow what the esptool does.

igrr (approved):
I've tried this locally and flashing did happen faster than before, LGTM

Issue #30, opened 2021-12-08
consider re-implemented the duplicated stdout without `multiprocessing` (RDT-80)


hfudev:
`multiprocessing` behaves differently on windows, macos, and linux. For now we could workaround this by using `fork` inside macos and linux, and `spawn` in windows. but in macos, `fork` is NOT recommended since it would cause error: https://bugs.python.org/issue33725

consider re-implement with https://pexpect.readthedocs.io/en/stable/api/popen_spawn.html#popenspawn-class and pass the stdout of other popen objects into the pexpect popen's stdin

(just an idea, need to investigate)

hfudev:
The reason that we couldn't use `pexpect.spawn` class anymore is that the stdin is a fake file descriptor and has no `fileno()`, counldn't be passed as `stdout` in other Popen objects.

hfudev:
thinking about supporting windows as well. but is really hard... fdexpect seems like a correct solution, but windows poorly support file descriptor stuff... Now the only working related stuff is `os.pipe()`, but the `os.read()` would be blocked if nothing's there. Looking for a solution of non-blocking pipe in windows...

hfudev:
https://docs.microsoft.com/en-us/windows/win32/ipc/named-pipe-type-read-and-wait-modes?redirectedfrom=MSDN

PR #31, opened 2021-12-10
refactor: new pexpect process approach with `os.pipe`

hfudev:
related https://github.com/espressif/pytest-embedded/issues/30

pexpect only support `fdspawn` and `PopenSpawn` on windows, and the we couldn't interact with a `PopenSpawn` process. That makes us only have one option `fdspawn`

We need to find a file descriptor like stuff that support read/write simouteniously and has real fileno (and support windows). seems that the `os.pipe` is the only option left.

The implementation of reading from the pipe in `pexpect` is here: https://github.com/pexpect/pexpect/blob/master/pexpect/spawnbase.py#L169. The `os.read` method would block everything and it never stops.

Our current approach is to do the `os.read(fd)` with a separate thread, and kill the main process at the very end. but this would make the rest tests and all the pytest session finished hook functions unreachable. 

pytest-xdist supports run each test case in a subprocess. would give it a try


hfudev:
use https://docs.python.org/3/library/threading.html#threading.Thread.daemon to exit the thread!!!

now it's working on all os! started to re-implement the other stuffs!

hfudev:
no this will not work. The file handle in windows is not inheritted into the child processes. Moving to #32

PR #32, opened 2021-12-15
feature: support windows

### LINKED ISSUE #30: consider re-implemented the duplicated stdout without `multiprocessing` (RDT-80)

hfudev:
closes #30


hfudev:
@igrr @tomassebestik should we have some windows/macos runners to test our platform compatibility?

since this plugin would be the main test framework in IDF repo in future I think it's important to support all os.

igrr:
Github actions have both Windows and macOS runners, i think we can run at least some test which doesn't require the serial port, such as `examples/echo`.

PR #33, opened 2021-12-15
Fix few typos in different places.

Ouss4:


hfudev:
@Ouss4 Thank you for your help :+1:

PR #35, opened 2021-12-21
Don't globally ignore build directory.

### LINKED ISSUE #34: build/ folder is not tracked. (RDT-82)
The `build` folder [is ignored](https://github.com/espressif/pytest-embedded/blob/main/.gitignore#L76),  however we do track a build folder under [tests/fixtures/hello_world_esp32/build](https://github.com/espressif/pytest-embedded/tree/main/tests/fixtures/hello_world_esp32/build).  Was the intentent to ignore folders generated when building the examples at [examples/esp-idf](https://github.com/espressif/pytest-embedded/tree/main/examples/esp-idf)?  If that's the case, we can have a `.gitignore` at `esp-idf` or `examples`, this way we ignore those generated folders but we still can use `build/` elsewhere.

Context is, I need to add a `build` folder under `tests/fixtures/` for the Arduino service tests.

Ouss4:
Closes #34

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,11 +69,3 @@ venv/
 
 # IDE
 .idea/
-
-# idf related
-sdkconfig
-sdkconfig.old
-build/
-
-# qemu image
-flash_image.bin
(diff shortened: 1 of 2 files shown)


hfudev:
Thank you :)

PR #37, opened 2021-12-22
Add Arduino service. (RDT-84)

Ouss4:
This PR adds an Arduino service to be able to parse an example, flash it and run tests.


Ouss4:
@hfudev I added pre-built binaries under `tests/fixtures` for an Arduino example.  I don't know if you want to keep that.  Are we going to add other binaires for other future services?
If you only want to keep IDF's binaries for one example let me know, I'll delete them and skip the Arduino service tests.

pytest-embedded-arduino/setup.py
```diff
@@ -0,0 +1,61 @@
...
+    return codecs.open(file_path, encoding='utf-8').read()
+
+
+def get_version():
+    regex = re.compile(r'^version = "([0-9.]+)"$', re.MULTILINE)
+    return regex.findall(read('../pyproject.toml'))[0]
+
+
+AUTHOR = 'Fu Hanxi'
+EMAIL = 'fuhanxi@espressif.com'
```
hfudev, line 20: feel free to change the author name and email :)

.gitignore
```diff
@@ -69,11 +69,3 @@ venv/
 
 # IDE
 .idea/
-
```
hfudev, line 72: these two ignore files are merged. please rebase

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -349,9 +350,15 @@ def pytest_addoption(parser):
         '- idf: auto-detect more app info with idf specific rules, auto flash-in\n'
         '- jtag: openocd and gdb\n'
         '- qemu: use qemu simulator instead of the real target\n'
+        '- arduino: auto-detect more app info with arduino specific rules, auto flash-in\n'
         'All the related CLI options are under the groups named by "embedded-<service>"',
     )
     base_group.addoption('--app-path', help='App path')
+    base_group.addoption('--build-dir', help='build directory under the app_path. (Default: "build")')
```
hfudev, line 357: `build_dir` move to base `App` is a good idea. However I think we need to add this argument to all `App` and its sub-classes. Now the base `App` doesn't have this attribute.

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -349,9 +350,15 @@ def pytest_addoption(parser):
...
+    base_group.addoption(
```
hfudev, line 358: I think `esp_group` is a better place to go. we could move the attribute to `EspSerial` as well.

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -682,6 +684,17 @@ def _fixture_classes_and_options(
                             'part_tool': part_tool,
                         }
                     )
+            elif 'arduino' in _services:
+                from pytest_embedded_arduino.app import ArduinoApp
+
+                classes[fixture] = ArduinoApp
+                kwargs[fixture].update(
+                    {
+                        'pexpect_proc': pexpect_proc,
```
hfudev, line 693: After moving `build_dir` to the base `App` class, we could move these `kwargs` all to the default one

```python
    if fixture == 'app':
        kwargs['app'] = {'app_path': app_path, 'build_dir': build_dir}
```

hfudev (commented):
LGTM!

hfudev:
@Ouss4 Nice work! For the binaries, I think we should keep them to pass the CI. and overall the Arduino ones are less than 500 kb. Let's keep them in the repo.

hfudev:
besides, I'm using https://www.conventionalcommits.org/en/v1.0.0/ to generate the changelog in this repo. Don't need to be super strict to it, but for the release notes related commits, please follow this to generate better messages.

Ouss4:
> besides, I'm using https://www.conventionalcommits.org/en/v1.0.0/ to generate the changelog in this repo. Don't need to be super strict to it, but for the release notes related commits, please follow this to generate better messages.

Yes, I read the contributing guide and tried to follow the same thing.  However, I used that only for the main commit, I thought that would be enough to add to the changelog.

Should we do the same for all commits?  Wouldn't that crowd the changelog?

hfudev:
the tool would ignore those commits starts with `ci: ...` or `example: ...`. and keep those with `feat: ...`, `fix: ...`.

but yes, you're right. we don't have to do that for all commits. it's nice to follow but not a strict rule!

Ouss4:
> the tool would ignore those commits starts with ci: ... or example: .... and keep those with feat: ..., fix: ....

Ah I see.  I thought it would pick everything.  So, I will keep `feat:` for the main commit and use something else for the rest.

pytest-embedded/pytest_embedded/plugin.py
Ouss4: Here you are talking about the `skip_autflash` attribute, right?

hfudev: yes.

Ouss4:
@hfudev I addressed your comments, PTAL agian.

pytest-embedded/pytest_embedded/app.py
```diff
@@ -7,14 +7,22 @@ class App:
...
+    def __init__(
+        self,
+        app_path: Optional[str] = None,
+        build_dir: Optional[str] = None,
+        **kwargs,
+    ):
         """
         Args:
             app_path: App path
+            build_dr: Build directory (where binaries reside)
```
hfudev, line 19: build_dir

hfudev (commented):
only one nitpick. let's merge it after this tiny fix.

pytest-embedded/pytest_embedded/app.py
Ouss4: Thanks, fixed!

Issue #29, opened 2021-12-08
bring the qemu stderr back to the pexpect process (RDT-79)


hfudev:
now there's no stderr back. only bring the serial output back as a serial file

igrr:
I also noticed that since we pass `-serial file:{self.log_file}` to QEMU, `dut.write` for QemuDut also doesn't work. This is an issue if the test app expects us to choose the test to run, like the unit-test-app does.

PR #41, opened 2022-01-10
fix(qemu): `dut.write` to qemu process correctly

hfudev:
related #29


igrr (approved):
LGTM, thanks for the fix!

PR #68, opened 2022-03-10
pytest_embedded_arduino/app.py: Add ESP32-S3 to the list of supported chips (RDT-107)

Ouss4:
Include ESP32-S3 to the list of supported chips in Arduino.


igrr:
LGTM, just one general question. These offsets will probably depend on the specific partition table selected in Tools menu? (Or in the FQBN string if building via arduino-builder.) Do you see a way how we can avoid hardcoding these offsets in pytest-embedded and pull this information from arduino?

In IDF there is a "flasher_args" JSON file produced by the build system, which includes flashing offsets. I wonder if we need to generate something similar in Arduino as well.

Ouss4:
> In IDF there is a "flasher_args" JSON file produced by the build system, which includes flashing offsets. I wonder if we need to generate something similar in Arduino as well.

That would be ideal.  Otherwise we'll have to do the same as the Arduino build system to extract the partition table from the FQBN.
Arduino build system generates a `build.options.json` file that includes the FQBN but not the offsets. Not sure how to do that for now.
cc. @me-no-dev

me-no-dev:
we don't even do it in Arduino currently, but builder should have copied the selected partition scheme to `[test]/build/partitions.csv`. If you need to change something, please do it on the `esp32-s3-support` branch.

Ouss4:
> we don't even do it in Arduino currently, but builder should have copied the selected partition scheme to `[test]/build/partitions.csv`. If you need to change something, please do it on the `esp32-s3-support` branch.

We'll think about that after this one is done.

@hfudev is the failing test supposed to pass?  Python 3.10 is passing but not 3.6.

hfudev:
@Ouss4 

> @hfudev is the failing test supposed to pass? Python 3.10 is passing but not 3.6.

I think it's related to #69.

Let's merge this when you think this MR is ready.

Ouss4:
@hfudev thanks for the explanation.  This is one is ready to be merged.

PR #74, opened 2022-03-18
Fix stub loader can never read serial issue

hfudev:
the redirect_io_thread is always reading from serial, need to disconnect and reconnect first!


hfudev:
this is a regression from #72 

will add a test case about it

PR #77, opened 2022-03-24
Added get_sha256 function to get SHA256 (RDT-181)

hmalpani:
This PR adds `get_sha256` function which is used to get SHA256 of app

### DIFF
--- a/pytest-embedded-idf/pytest_embedded_idf/app.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/app.py
@@ -188,3 +188,17 @@ class IdfApp(App):
         flash_settings['encrypt'] = any([file.encrypted for file in flash_files])
 
         return flash_args, flash_files, flash_settings
+
+    def get_sha256(self, path_to_file):
+        """
+        get sha256 of app
+
+        :param: path: path to file
+        :return: sha256 appended to app
+        """
+        from esptool import LoadFirmwareImage, hexify
+
+        image = LoadFirmwareImage(self.target, path_to_file)
+        if image.append_digest:
+            return hexify(image.stored_digest).lower()
+        return None


pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -35,3 +48,31 @@ def toolchain_prefix(self) -> str:
             return f'riscv32-{self.target}-elf-'
         else:
             raise ValueError(f'Unknown target: {self.target}')
+
+    def image_info(self, path_to_file):
```
igrr, line 52: Should this be a method of `App` rather than the `Dut`?
Looks like this method doesn't use anything from `IdfDut`, and is not dependent on even having a Dut instance?

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -35,3 +48,31 @@ def toolchain_prefix(self) -> str:
...
+        """
+        get hash256 of app
+
+        :param: path: path to file
+        :return: sha256 appended to app
+        """
+
+        old_stdout = sys.stdout
+        new_stdout = io.StringIO()
+        sys.stdout = new_stdout
```
igrr, line 62: If an exception happens between this line and `sys.stdout = old_stdout` below, sys.stdout won't be restored

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -35,3 +48,31 @@ def toolchain_prefix(self) -> str:
...
+                    self.__setattr__(key, value)
+
+        args = Args(
+            {
+                'chip': self.target,
+                'filename': path_to_file,
+            }
+        )
+        esptool.image_info(args)
+        output = new_stdout.getvalue()
```
igrr, line 76: Does this return only the sha256 hash appended to the app? When I run `esptool.py image_info`, i get multiple lines of output instead:
```
esptool.py v3.3-dev
Image version: 1
Entry point: 400810d4
6 segments

Segment 1: len 0x07764 load 0x3f400020 file_offs 0x00000018 [DROM]
Segment 2: len 0x02674 load 0x3ffb0000 file_offs 0x00007784 [BYTE_ACCESSIBLE,DRAM]
Segment 3: len 0x06210 load 0x40080000 file_offs 0x00009e00 [IRAM]
Segment 4: len 0x146b0 load 0x400d0020 file_offs 0x00010018 [IROM]
Segment 5: len 0x05600 load 0x40086210 file_offs 0x000246d0 [IRAM]
Segment 6: len 0x00010 load 0x50000000 file_offs 0x00029cd8 [RTC_DATA]
Checksum: f4 (valid)
Validation Hash: d0cd46e5d6ba47189cb1360468d5e863fee9b91e52dff33d89a1f5d7b72fe135 (valid)
```

Looking at esptool code, we could possibly get just the SHA256 hash appended to the file using something like
```python
image = esptool.LoadFirmwareImage(self.target, path_to_file)
if image.append_digest:
    return image.stored_digest
```

However I don't know if we can rely on these esptool internals to be stable. OTOH, I am not sure we can rely on image_info output format to be stable, either. @radimkarnis what would you suggest?

pytest-embedded-idf/pytest_embedded_idf/dut.py
hmalpani: It uses `target` from `IdfDut`

pytest-embedded-idf/pytest_embedded_idf/dut.py
hmalpani: > Does this return only the sha256 hash appended to the app? When I run `esptool.py image_info`, i get multiple lines of output instead:
> 
> ```
> esptool.py v3.3-dev
> Image version: 1
> Entry point: 400810d4
> 6 segments
> 
> Segment 1: len 0x07764 load 0x3f400020 file_offs 0x00000018 [DROM]
> Segment 2: len 0x02674 load 0x3ffb0000 file_offs 0x00007784 [BYTE_ACCESSIBLE,DRAM]
> Segment 3: len 0x06210 load 0x40080000 file_offs 0x00009e00 [IRAM]
> Segment 4: len 0x146b0 load 0x400d0020 file_offs 0x00010018 [IROM]
> Segment 5: len 0x05600 load 0x40086210 file_offs 0x000246d0 [IRAM]
> Segment 6: len 0x00010 load 0x50000000 file_offs 0x00029cd8 [RTC_DATA]
> Checksum: f4 (valid)
> Validation Hash: d0cd46e5d6ba47189cb1360468d5e863fee9b91e52dff33d89a1f5d7b72fe135 (valid)
> ```
This output is correct. Just saw the `simpla_ota_example_test` and found that sha256 is searched in this output.


> However I don't know if we can rely on these esptool internals to be stable. OTOH, I am not sure we can rely on image_info output format to be stable, either. @radimkarnis what would you suggest?

This function is the same as in `ttfw_idf`.

igrr: The thing is, ttfw_idf is kept in the IDF repository, where esptool dependency is expressed using a submodule. If esptool API changes, we will simply update ttfw_idf in the MR where esptool is updated.

However pytest_embedded version is not linked to a specific esptool version, which means that it should (at least try to) be forward and backward compatible. Let's see what esptool folks say.

pytest-embedded-idf/pytest_embedded_idf/dut.py
igrr: IdfApp also has the .target property, right?

I'm just thinking that getting the SHA256 hash attached to the binary image seems to be more like a utility function related to the app binary image. It doesn't look related to the DUT. Besides, IdfApp also "knows" the app binary path, so the benefit is that it doesn't need to be passed as an argument.

pytest-embedded-idf/pytest_embedded_idf/dut.py
igrr: > This output is correct. Just saw the simpla_ota_example_test and found that sha256 is searched in this output.

Does it make sense to either rename the function and update the function description here (currently it claims to return just the sha256) or to make it return just the sha256, rather than the whole image_info output?

radimkarnis: This approach LGTM! the API will soon change to:
```python
image = esptool.bin_image.LoadFirmwareImage(self.target, path_to_file)
```
Other than that, the image_info output can be considered stable, there are no plans to change it.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -1,9 +1,22 @@
...
+import os
+import sys
+
 from pytest_embedded import PexpectProcess
 from pytest_embedded_serial.dut import SerialDut
 
 from .app import IdfApp
 from .serial import IdfSerial
 
+try:
```
hfudev, line 11: I think a simple `import esptool` is safe here. The setup.py would install the esptool package.

hfudev: The `IdfDut` would only be used when `esp, idf` services are all enabled

```python
if 'esp' in _services and 'idf' in _services:
    from pytest_embedded_idf.dut import IdfDut

    classes[fixture] = IdfDut
```

pytest-embedded-idf/pytest_embedded_idf/dut.py
hfudev: I think this is okay to move to IdfApp. but since IdfApp is not 100% guaranteed includes esptool when using, (for example it is used alone without service `esp`), I recommend the `import esptool` to be put inside the function itself.

hfudev: and we don't need a try except ImportError by this approach.

pytest-embedded-idf/pytest_embedded_idf/dut.py
hfudev: if the test requires to calculate the sha256 to a bin file, I think it's better to do it on the test script side, not adding a function here. the esptool call is not needed.

igrr: @hfudev it's not exactly the sha256 of the bin file. It's SHA256 of most of the bin file, except for the SHA256 itself, which is appended at the end of the bin file :)

https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/firmware-image-format.html#footer

(it is also optional, so some bin files might not have this hash appended)

hfudev: Now I get it. Thanks!

hmalpani:
@hfudev I have updated the PR. PTAL

hfudev:
Thanks. LGTM!

hfudev (approved):

PR #93, opened 2022-06-02
Fix esptool imports (RDT-222)

hmalpani:
This MR fixes esptool imports.

https://github.com/espressif/pytest-embedded/pull/77#discussion_r834197044

### DIFF
--- a/pytest-embedded-idf/pytest_embedded_idf/app.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/app.py
@@ -214,7 +214,13 @@ class IdfApp(App):
         Returns:
             sha256 appended to app
         """
-        from esptool import LoadFirmwareImage, hexify
+        from pytest_embedded_serial_esp.serial import EspSerial, EsptoolVersion
+
+        if EspSerial.ESPTOOL_VERSION == EsptoolVersion.V3:
+            from esptool import LoadFirmwareImage, hexify
+        else:
+            from esptool.bin_image import LoadFirmwareImage
+            from esptool.util import hexify
 
         image = LoadFirmwareImage(self.target, filepath)
         if image.append_digest:


hmalpani:
@hfudev PTAL

pytest-embedded-idf/pytest_embedded_idf/app.py
```diff
@@ -214,7 +214,8 @@ def get_sha256(self, filepath: str) -> Optional[str]:
         Returns:
             sha256 appended to app
         """
-        from esptool import LoadFirmwareImage, hexify
+        from esptool.bin_image import LoadFirmwareImage
+        from esptool.util import hexify
```
hfudev, line 218: could you use the `self.ESPTOOL_VERSION == EsptoolVersion.V3` or `self.ESPTOOL_VERSION == EsptoolVersion.V4` to determine the correct import here? We still need to make the code compatible with esptool 3.x.

hfudev (changes requested):

pytest-embedded-idf/pytest_embedded_idf/app.py
hmalpani: Updated the PR

pytest-embedded-idf/pytest_embedded_idf/app.py
```diff
@@ -214,7 +214,15 @@ def get_sha256(self, filepath: str) -> Optional[str]:
         Returns:
             sha256 appended to app
         """
-        from esptool import LoadFirmwareImage, hexify
+        from pytest_embedded_serial_esp.serial import EsptoolVersion
+
+        from .serial import IdfSerial
```
hfudev, line 219: sorry to be picky...

```python
from pytest_embedded_serial_esp.serial import EspSerial, EsptoolVersion

if EspSerial.ESPTOOL_VERSION == ...
```

would be simpler.

hfudev (changes requested):

PR #98, opened 2022-06-29
Add erase-flash option to erase the flash before programming. (RDT-234)

Ouss4:
Add an option to erase the flash before programming. And example that needs this is: https://github.com/espressif/arduino-esp32/pull/6916
Tested with the Arduino example by adding:

```diff
diff --git a/examples/arduino/hello_world/test_hello_world.py b/examples/arduino/hello_world/test_hello_world.py
index 0e7cea8..03f7346 100644
--- a/examples/arduino/hello_world/test_hello_world.py
+++ b/examples/arduino/hello_world/test_hello_world.py
@@ -1,2 +1,9 @@
+import pytest
+
+@pytest.mark.parametrize(
+    'erase_flash',
+    ['True'],
+    indirect=True,
+)
 def test_hello_arduino(dut):
     dut.expect('Hello Arduino!')
```

IDF service added a similar function bu I am not sure how that one is used.

### DIFF
--- a/pytest-embedded-arduino/pytest_embedded_arduino/serial.py
+++ b/pytest-embedded-arduino/pytest_embedded_arduino/serial.py
@@ -25,10 +25,11 @@ class ArduinoSerial(EspSerial):
         baud: int = EspSerial.DEFAULT_BAUDRATE,
         target: Optional[str] = None,
         skip_autoflash: bool = False,
+        erase_flash: bool = False,
         **kwargs,
     ) -> None:
         self.app = app
-        super().__init__(pexpect_proc, target or self.app.target, port, baud, skip_autoflash, **kwargs)
+        super().__init__(pexpect_proc, target or self.app.target, port, baud, skip_autoflash, erase_flash, **kwargs)
 
     def _start(self):
         if self.skip_autoflash:
@@ -67,6 +68,9 @@ class ArduinoSerial(EspSerial):
         if self.ESPTOOL_VERSION == EsptoolVersion.V4:
             default_kwargs['force'] = False
             default_kwargs['chip'] = self.app.target
+
+        if self.erase_flash:
+            default_kwargs['erase_all'] = True
 
         default_kwargs.update(self.app.flash_settings)
         flash_args = FlashArgs(default_kwargs)
--- a/pytest-embedded-idf/pytest_embedded_idf/serial.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/serial.py
@@ -29,6 +29,7 @@ class IdfSerial(EspSerial):
         port: Optional[str] = None,
         baud: int = EspSerial.DEFAULT_BAUDRATE,
         skip_autoflash: bool = False,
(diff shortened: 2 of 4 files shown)


hfudev:
Hi @Ouss4! if I understand correctly, you're trying to erase the nvs right? (guess based on [this](https://github.com/espressif/arduino-esp32/pull/6885#issuecomment-1167220819)). We already have a `erase_nvs` field for idf. could you use integrate `erase_nvs` to arduino?

Ouss4:
> Hi @Ouss4! if I understand correctly, you're trying to erase the nvs right? (guess based on [this](https://github.com/espressif/arduino-esp32/pull/6885#issuecomment-1167220819)). We already have a `erase_nvs` field for idf. could you use integrate `erase_nvs` to arduino? (instead add another cli option `erase_all`)

Hi,
Well, that tests requires to only erase the NVS partition but it has been requested to erase the whole flash too. So I added this.

Ouss4:
BTW, @hfudev we do have a function in IDF service that erases the flash but it's not exposed, do you know where that one is used?  I know it wasn't you who added it.

hfudev:
> BTW, @hfudev we do have a function in IDF service that erases the flash but it's not exposed, do you know where that one is used? I know it wasn't you who added it.

like here: https://github.com/espressif/esp-idf/blob/master/examples/system/ota/advanced_https_ota/pytest_advanced_ota.py#L473

pytest-embedded-arduino/pytest_embedded_arduino/serial.py
```diff
@@ -68,6 +69,9 @@ def __init__(self, attributes):
             default_kwargs['force'] = False
             default_kwargs['chip'] = self.app.target
 
+        if self.erase_flash:
+            default_kwargs['erase_all'] = True
```
hfudev, line 73: here should be `default_kwargs['erase_flash']`?

hfudev (changes requested):
@Ouss4 LGTM! It seems useful. Do you mind add this cli option for idf service as well?

pytest-embedded-arduino/pytest_embedded_arduino/serial.py
Ouss4: This is the argument passed to `esptool`, it's called `erase_all`.  I tried with `erase_flash` to keep it consistent, but that's not recognized.

hfudev: ah, sorry. my mistake.

hfudev:
> > Do you mind add this cli option for idf service as well?
> 
> I added it to IDF service. What to do with the function `erase_flash`? Should we delete it and use the parameter? This will require tests that use that function to be updated.

let's keep the old `erase_flash` function, since if we remove this it would requires changing the test scripts in IDF.

hfudev (approved):

hfudev:
Thank you @Ouss4 !

PR #99, opened 2022-06-29
foreach.sh: Update the script

Ouss4:
1. Add the bash shebang
  2. Add the rest of the usual options that go with set -e (`set -euo pipefail`)
  3. make the file executable.

### DIFF
--- a/foreach.sh
+++ b/foreach.sh
@@ -1,4 +1,6 @@
-set -eo
+#!/bin/bash
+
+set -euo pipefail
 
 DEFAULT_PACKAGES=" \
   pytest-embedded \


foreach.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 set -eo
```
igrr, line 3: Since you are changing this, could you please also modify this line to `set -euo pipefail`?

1. that adds `-u`, which catches uninitialized variables
2. `set -o` simply prints the values of shell options, i don't think it's what was intended here. OTOH, `set -o pipefail` is commonly added to `set -eu`.

Ouss4: Thanks, done.

hfudev:
thanks!

PR #103, opened 2022-07-11
fix cli option "--erase-flash" conflict with function `erase_flash()`

hfudev:
fix regression from https://github.com/espressif/pytest-embedded/pull/98


hfudev:
@Ouss4 PTAL.

hfudev:
yanked 0.7.4

Ouss4 (approved):
Ah so it ended up conflicting with the function... I did have concerns that this may happen.
Thanks for the fix @hfudev.

hfudev:
added some tests

Issue #96, opened 2022-06-23
Make Serial Settings Configurable


hfudev:
As  #95 and #39 reported, we need a better way to configure the serial port from the cli arguments.

For now we only provide workarounds for baud rate as env var `ESPBAUD`

PR #107, opened 2022-07-18
Feat/baud fixture move to serial service

hfudev:
related #96

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -67,8 +67,8 @@ embedded:
                         "--app-path=test_path1|test_path2" when two DUTs are using different built binary files.
                         "--part-tool=part_tool_path|" when only the first DUT needs this option, the second should keep as empty.
                         "--embedded-services=idf --count=2" when both of these DUTs are using the same services.
-                        The configuration would be duplicated when it has only one value but the "count" amount is greater than 1. It would raise an
-                        exception when the configuration has multi values but the amount is different from the "count" amount.
+                        The configuration would be duplicated when it has only one value but the "count" amount is greater than 1. It would raise an exception when the
+                        configuration has multi values but the amount is different from the "count" amount.
                         For example:
                         "--embedded-services=idf|esp-idf --count=3" would raise an exception.
   --parallel-count=PARALLEL_COUNT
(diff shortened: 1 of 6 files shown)


docs/services.md
```diff
@@ -88,19 +88,34 @@ embedded:
...
   --build-dir=BUILD_DIR
                         build directory under the app_path. (Default: "build")
+  --with-timestamp=WITH_TIMESTAMP
+                        y/yes/true for True and n/no/false for False. Set to True to enable print with timestamp. (Default: True)
+  --reorder-by-app-path
+                        Reorder the test sequence according to the [app_path] and [build_dir]. (Default: False)
 
 embedded-serial:
   --port=PORT           serial port. (Env: "ESPPORT" if service "esp" specified, Default: "None")
+  --baud=BAUD           serial port baud rate used when flashing. (Env: "ESPBAUD" if service "esp" specified, Default: 115200)
```
igrr, line 98: It's a bit odd here since "flashing" is something related to `esp` service. Other microcontrollers don't usually have a concept of "flashing" over serial port.

The other problem is that now this argument controls both flashing baud rate (for `esp`) and monitor baud rate (for `serial`). A common use case is to set flashing baud rate higher (to reduce flashing time) but still use standard 115200 baud rate for monitoring, because that's the baud rate the application uses.

igrr: I would suggest to:

- keep `--baud` option here (in `serial`), change the description to say that it sets the monitoring/communication baud rate. Remove the mention of ESPBAUD and flashing.
- add another option, let's say `--esptool-baud`, in `esp` service. It will be set from ESPBAUD by default, and the description of the option will be related to flashing.

hfudev: and do you think I should keep this "auto change flash baud rate" mechanism?

igrr: What about setting the default value of --esptool-baud argument to 921600? Then no extra logic is necessary. The user can either raise of lower the flashing baudrate by passing this argument.
(They may want to lower the baud rate if their hardware doesn't work well at 921600 baud, for example.)

PR #109, opened 2022-07-21
feat(idf): add function to parse case list from unit test menu (RDT-243)

wangyz1997:
Add a function to get test case list from test menu via UART print. 

Input:
```
(54)	"sdio interrupt" [sdio][test_env=UT_SDIO][multi_device]
	(1)	"test_sdio_interrupt_master"
	(2)	"test_sdio_interrupt_slave"
(55)	"sdio register" [sdio][test_env=UT_SDIO][multi_device]
	(1)	"test_sdio_reg_master"
	(2)	"test_sdio_interrupt_slave"
(56)	"sdio reset" [sdio][test_env=UT_SDIO][multi_device]
	(1)	"test_sdio_reset_master"
	(2)	"test_sdio_reset_slave"
(57)	"master slave test: SDIO_FRHOST" [sdio][timeout=180][test_env=UT_SDIO][multi_device]
	(1)	"PTEST_MASTER_SDIO_FRHOST"
	(2)	"PTEST_SLAVE_SDIO_FRHOST"
(58)	"master slave test: SDIO_TOHOST" [sdio][timeout=180][test_env=UT_SDIO][multi_device]
	(1)	"PTEST_MASTER_SDIO_TOHOST"
	(2)	"PTEST_SLAVE_SDIO_TOHOST"
```

Output:
```
UnittestMenuCase(index=54, name='sdio interrupt', type='multi_device', keywords=[], groups=['sdio'], attributes={'test_env': 'UT_SDIO'}, subcases=[{'index': '1', 'name': 'test_sdio_interrupt_master'}, {'index': '2', 'name': 'test_sdio_interrupt_slave'}])
UnittestMenuCase(index=55, name='sdio register', type='multi_device', keywords=[], groups=['sdio'], attributes={'test_env': 'UT_SDIO'}, subcases=[{'index': '1', 'name': 'test_sdio_reg_master'}, {'index': '2', 'name': 'test_sdio_interrupt_slave'}])
UnittestMenuCase(index=56, name='sdio reset', type='multi_device', keywords=[], groups=['sdio'], attributes={'test_env': 'UT_SDIO'}, subcases=[{'index': '1', 'name': 'test_sdio_reset_master'}, {'index': '2', 'name': 'test_sdio_reset_slave'}])
UnittestMenuCase(index=57, name='master slave test: SDIO_FRHOST', type='multi_device', keywords=[], groups=['sdio'], attributes={'timeout': '180', 'test_env': 'UT_SDIO'}, subcases=[{'index': '1', 'name': 'PTEST_MASTER_SDIO_FRHOST'}, {'index': '2', 'name': 'PTEST_SLAVE_SDIO_FRHOST'}])
UnittestMenuCase(index=58, name='master slave test: SDIO_TOHOST', type='multi_device', keywords=[], groups=['sdio'], attributes={'timeout': '180', 'test_env': 'UT_SDIO'}, subcases=[{'index': '1', 'name': 'PTEST_MASTER_SDIO_TOHOST'}, {'index': '2', 'name': 'PTEST_SLAVE_SDIO_TOHOST'}])
```

### DIFF
--- a/pytest-embedded-idf/pytest_embedded_idf/dut.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/dut.py
@@ -3,12 +3,38 @@ import os
 import re
 import tempfile
 from contextlib import redirect_stdout
+from dataclasses import dataclass
+from typing import List
 
 from pytest_embedded import PexpectProcess
 from pytest_embedded_serial.dut import SerialDut
 
 from .app import IdfApp
 from .serial import IdfSerial
+
+
+@dataclass
+class UnittestMenuCase:
+    """
+    Dataclass of esp-idf unit test cases parsed from test menu. This datachass has following fields:
+
+        - index: The index of the case, which can be used to run this case.
+        - name: The name of the case.
+        - type: Type of this case, which can be `normal` `multi_stage` or `multi_device`.
+        - keywords: List of additonal keywords of this case. For now, we have `disable` and `ignored`.
+        - groups: List of groups of this case, this is usually the component which this case belongs to.
+        - attributes: List of attributes of this case, which is used to describe timeout duration,
+            test environment, etc.
+        - subcases: List of subcases of this case, if this case is a `multi_stage` or `multi_device` one.
+    """
+
+    index: int
+    name: str
+    type: str
+    keywords: List
+    groups: List
+    attributes: List
+    subcases: List
 
 
 class IdfDut(SerialDut):
@@ -127,3 +153,100 @@ class IdfDut(SerialDut):
         if not self.skip_check_coredump:
             self._check_coredump()
         super().close()
+
+    def parse_test_menu(
+        self,
+        ready_line: str = 'Press ENTER to see the list of tests',
+        pattern="Here's the test menu, pick your combo:(.+)Enter test for running.",
+        trigger: str = '',
+    ) -> List[UnittestMenuCase]:
+        """
+        Get test case list from test menu via UART print.
+
+        Args:
+            ready_line: Prompt to indicate that device is ready to print test menu.
+            pattern: Pattern to match the output from device, menu block should be in the first group.
+                     This will be directly passed to `pexpect.expect()`.
+            trigger: Keys to trigger device to print test menu by UART.
+
+        Returns:
+            A `list` of `UnittestMenuCase`, which includes info for each test case.
+        """
+        self.expect_exact(ready_line)
+        self.write(trigger)
+        menu_block = self.expect(pattern).group(1)
+        s = str(menu_block, encoding='UTF-8')
+        return self.parse_unity_menu_from_str(s)
+
+    @staticmethod
+    def parse_unity_menu_from_str(s: str) -> List[UnittestMenuCase]:
+        """
+        Parse test case mcnu from string to list of `UnittestMenuCase`.
+
+        Args:
+            s: string include test case menu.
+
+        Returns:
+            A `list` of `UnittestMenuCase`, which includes info for each test case.
+        """
+        print(s)
+        cases = s.splitlines()
+
+        case_regex = re.compile(r'\((\d+)\)\s\"(.+)\"\s(\[.+\])+')
+        subcase_regex = re.compile(r'\t\((\d+)\)\s\"(.+)\"')
+
+        test_menu = []
+        for case in cases:
+            case_match = case_regex.match(case)
+            if case_match is not None:
+                index, name, tag_block = case_match.groups()
+                tags = re.findall(r'\[(.+?)\]', tag_block)
+
+                if 'multi_stage' in tags:
+                    type = 'multi_stage'
+                    tags.remove('multi_stage')
+                elif 'multi_device' in tags:
+                    type = 'multi_device'
+                    tags.remove('multi_device')
+                else:
+                    type = 'normal'
+
+                keyword = []
+                if 'ignore' in tags:
+                    keyword.append('ignore')
+                    tags.remove('ignore')
+                elif 'disable' in tags:
+                    keyword = 'disable'
+                    tags.remove('disable')
+
+                attribute = {}
+                group = []
+                for tag in tags:
+                    if '=' in tag:
+                        k, v = tag.replace(' ', '').split('=')
+                        attribute[k] = v
+                    else:
+                        group.append(tag)
+
+                test_menu.append(
+                    UnittestMenuCase(
+                        index=int(index),
+                        name=name,
+                        type=type,
+                        keywords=keyword,
+                        groups=group,
+                        attributes=attribute,
+                        subcases=[],
+                    )
+                )
+                continue
+            subcase_match = subcase_regex.match(case)
+            if subcase_match is not None:
+                index, name = subcase_match.groups()
+                test_menu[-1].subcases.append({'index': index, 'name': name})
+                continue
+
+            if case != '':
+                raise NotImplementedError('Unrecognized test case:', case)
+
+        return test_menu
--- a/pytest-embedded-idf/tests/test_idf.py
+++ b/pytest-embedded-idf/tests/test_idf.py
@@ -1,6 +1,8 @@
 import os
 import re
 
+from pytest_embedded_idf.dut import IdfDut
+
 
 def test_idf_serial_flash(testdir):
     testdir.makepyfile("""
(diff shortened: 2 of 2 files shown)


pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +128,80 @@ def close(self) -> None:
...
+    subcase_regex = re.compile(r'\t\((\d+)\)\s\"(.+)\"')
+
+    def get_test_menu(self) -> List:
+        """
+        Get test case list from test menu via UART print.
+
+        This function should be called after `Press ENTER to see the list of tests` prompt is shown.
+
+        Returns:
+            A `list` of avaliable test cases, in `dict`. Each test case's `dict` has following keys:
```
igrr, line 142: Could we use an instance of NamedTuple or a class instead of a `dict`? It's a bit more maintainable than a dictionary, plus it's better for documentation and for autocompletion hints in IDEs.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +128,80 @@ def close(self) -> None:
...
+                - name: The name of the case.
+                - type: Type of this case, which can be `normal` `multi_stage` or `multi_device`.
+                - keyword: List of additonal keywords of this case. For now, we have `disable` and `ignored`.
+                - group: List of groups of this case, this is usually the component which this case belongs to.
+                - attribute: List of attributes of this case, which is used to describe timeout duration,
+                    test environment, etc.
+                - subcase: List of subcases of this case, if this case is a `multi_stage` or `multi_device` one.
+        """
+        self.write('')
+        menu_block = self.expect("Here's the test menu, pick your combo:(.+)Enter test for running.").group(1)
```
igrr, line 153: Could we make the command to show the menu (here it is a newline) and the prompt message ("Here's the test menu...") configurable, for example as optional arguments to this function?

I have another test app which prints the list of tests after writing the command `test -h` and has a slightly different prompt.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +129,88 @@ def close(self) -> None:
         if not self.skip_check_coredump:
             self._check_coredump()
         super().close()
+
+    Case = namedtuple('Case', 'index name type keyword group attribute subcase')
```
hfudev, line 133: a list should be named with plurals.

```suggestion
    Case = namedtuple('Case', 'index name type keywords groups attributes subcases')
```

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +129,88 @@ def close(self) -> None:
...
+
+    def get_test_menu(
+        self, regex: str = "Here's the test menu, pick your combo:(.+)Enter test for running.", trigger: str = ''
```
hfudev, line 136: all the `expect` functions could have 3 data types: `str`, `re.Pattern` and list of other types. How about we keep the same API as the expect functions?

```suggestion
        self, pattern = "Here's the test menu, pick your combo:(.+)Enter test for running.", trigger: str = ''
```

pytest-embedded-idf/pytest_embedded_idf/dut.py
hfudev: or could use https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass to make your editor/IDE lint with data types.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +129,88 @@ def close(self) -> None:
...
+        menu_block = self.expect(regex).group(1)
+        cases = str(menu_block, encoding='UTF-8').splitlines()
+
+        case_regex = re.compile(r'\((\d+)\)\s\"(.+)\"\s(\[.+\])+')
+        subcase_regex = re.compile(r'\t\((\d+)\)\s\"(.+)\"')
+
+        test_menu = []
+        for case in cases:
+            case_match = case_regex.match(case)
+            if case_match is not None:
```
hfudev, line 168: here raise an exception when "if case_match is None and subcase_match is None" would make It easier to find those test menus with unexpected output.

hfudev (changes requested):
LGTM! only a few minor comments.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +129,88 @@ def close(self) -> None:
...
+
+    Case = namedtuple('Case', 'index name type keyword group attribute subcase')
+
+    def get_test_menu(
+        self, regex: str = "Here's the test menu, pick your combo:(.+)Enter test for running.", trigger: str = ''
+    ) -> List[Case]:
+        """
+        Get test case list from test menu via UART print.
+
+        This function should be called after `Press ENTER to see the list of tests` prompt is shown.
```
hfudev, line 141: How about moving this expect statement into this function as well, as an argument with a default value.

otherwise, a common test script would be 

```python
dut.expect('Press ENTER to see the list of tests')
menu_cases = dut.get_test_menu()
```

which is a bit redundant.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +129,88 @@ def close(self) -> None:
         if not self.skip_check_coredump:
             self._check_coredump()
         super().close()
+
+    Case = namedtuple('Case', 'index name type keyword group attribute subcase')
+
+    def get_test_menu(
```
hfudev, line 135: `get_xxx` is usually a simple get function. IMHO it would be better if the function is named `parse_test_menu`

hfudev (changes requested):

pytest-embedded-idf/pytest_embedded_idf/dut.py
wangyz1997: Test menu can be printed anytime when the device is idle, by pressing ENTER. So if we move expection of the prompt inside this function, we can only call this function right after test app launches. In fact, `Press ENTER to see the list of tests` is more like a signal of "device is ready" than "menu print is ready", so I think it's better to leave it in the top level of the test app to indicate it is ready for users to execute tests.

wangyz1997:
@hfudev PTAL.

pytest-embedded-idf/pytest_embedded_idf/dut.py
hfudev: For me, passing this as an argument with a default value is the same as giving the `trigger` as an argument with a default value.

Or you could split it up into two functions. One is `parse_unity_menu_from_str(s: str)`, which is a `staticmethod` that would parse the menu from a string. The other one is the current `parse_unity_menu`, which would have the `ready line`, `trigger`, and the `end pattern`.

By splitting it up into two functions, it would be easier to write a small test function without interacting with an actual test device. like https://github.com/espressif/pytest-embedded/blob/main/pytest-embedded/tests/test_base.py#L290

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +129,101 @@ def close(self) -> None:
...
+        This function should be called after `Press ENTER to see the list of tests` prompt is shown.
+
+        Args:
+            pattern: Pattern to match the output from chip, menu block should be in the first group.
+                     This will be directly passed to `pexpect.expect()`.
+            trigger: Keys to trigger chip to print test menu by UART.
+
+        Returns:
+            A `list` of avaliable test cases, in `Case`. `Case` has following fields:
+                - index: The index of the case, which can be used to run this case.
```
hfudev, line 158: It's better to move the descriptions into the class, and here could be simplified to "Returns: list of `Case`"

(this is to avoid modify two places if we're updating the class `Case`)

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +129,101 @@ def close(self) -> None:
         if not self.skip_check_coredump:
             self._check_coredump()
         super().close()
+
+    @dataclass
+    class Case:
```
hfudev, line 134: how about we change the name to `UnittestMenuCase`? The name `Case` would be a bit overbroad and is a widely used term in testing frameworks.

hfudev (changes requested):

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -127,3 +129,122 @@ def close(self) -> None:
...
+
+    @dataclass
+    class UnittestMenuCase:
+        """
+        Dataclass of esp-idf unit test cases parsed from test menu. This datachass has following fields:
+
+            - index: The index of the case, which can be used to run this case.
+            - name: The name of the case.
+            - type: Type of this case, which can be `normal` `multi_stage` or `multi_device`.
+            - keyword: List of additonal keywords of this case. For now, we have `disable` and `ignored`.
```
hfudev, line 141: `keywords`, same as the following ones

hfudev (commented):
LGTM! only one nitpick. Would you mind adding a small test case for the `parse_test_menu` function?

PR #110, opened 2022-07-22
only hook on dut of type pytest_embedded.dut.Dut

driftregion:
pytest docs say that fixtures can be overridden: 

https://docs.pytest.org/en/latest/how-to/fixtures.html?highlight=override#override-a-fixture-on-a-test-module-level

When I override the `dut` fixture in a module, it still gets caught by this `pytest_embedded` hook.

This change explicitly checks that the dut is of type `pytest_embedded.dut.Dut`


hfudev:
Thank you @driftregion!

PR #122, opened 2022-08-08
feat(idf): add test case runner for normal cases and multi-stage cases

wangyz1997:
This PR adds two functions in `IdfDut` class: `run_normal_cases` and `run_multi_stage_cases`.

The first function, `run_normal_cases`, implements a general method to run normal cases by passing a list of `UnittestMenuCase`. This function is now able to handle normal test cases with `timeout` `ignore` and `reset` tag. I've tried this function with `driver` and `esp_system` unit test, and the function is able to handle all the cases.

The second function, `run_multi_stage_cases`, implements a general method to run multi-stage cases by also passing a list of `UnittestMenuCase`. This function is now able to handle normal test cases with `timeout` `ignore` and `reset` tag as well. It's tested with all test cases in `esp_system` unit test and works out well.

For multi-device tests, it's hard to abstract a general model, since each case uses different `signals`, so function to run them are not included in this PR. User may extract multi-device tests from the test case list, and run them manually.

In this PR, I also changed the `ignore` list to `ignored` bool, which is used to indicate whether this case is ignored. I made this change because I found there is no case with `disable` label currently, and bool is easier for use. If there will be `disable` tags in the future, they will fall into `groups`. Also, I add a `strip` argument in `parse_test_menu`, which is true by default. If `strip` is true, the ignored cases will be striped from the returned list.

Here is an example on how to use those functions:

```
import pytest
from pytest_embedded import Dut

@pytest.mark.esp32
@pytest.mark.generic
def test_unit_test(dut: Dut) -> None:
    case_list = dut.parse_test_menu()
    dut.run_normal_cases([case for case in case_list if case.type == 'normal'])
    dut.run_multi_stage_cases([case for case in case_list if case.type == 'multi_stage'])
```


pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -156,6 +156,7 @@ def close(self) -> None:
 
     def parse_test_menu(
         self,
+        strip: bool = True,
```
hfudev, line 159: when adding an extra argument, it's better to add it at the very last with a default value to avoid breaking changes.

and please also improve the naming here. `strip` is more like a word for removing white spaces. how about `include_ignored` with a default value `False`?

hfudev (changes requested):
Overall LGTM. it's better to filter the test cases inside the real call function.

```python
import pytest
from pytest_embedded import Dut

@pytest.mark.esp32
@pytest.mark.generic
def test_unit_test(dut: Dut) -> None:
    case_list = dut.parse_test_menu()
    dut.run_normal_cases(case_list)
    dut.run_multi_stage_cases(case_list)
```

wangyz1997:
If we pass list of cases instead of single case into those runner functions, if user wants to run specific case, and a new list will need to be constructed, will it be more complicated?

Issue #87, opened 2022-05-10
[idf] Panic Handler Support (RDT-199)


hfudev:
Now we only support coredump from UART/Flash, parse raw panic output with https://github.com/espressif/esp-idf/blob/master/tools/gdb_panic_server.py

PR #123, opened 2022-08-16
feat: added panic handler support (RDT-265)

alekseiapa:
- added panic handler support

Closes https://github.com/espressif/pytest-embedded/issues/87


alekseiapa:
Hello @hfudev @igrr . Could you please take a look at the current PR ?

alekseiapa:
@hfudev Could you please help to run workflow ? Thanks !

alekseiapa:
@hfudev Could you please look over the current MR ?  There is one failed job that was caused by the Python environment. Thank you!

hfudev:
@alekseiapa Could you remove those unrelated files? normally these files should work

```
.
├── build
│   ├── bootloader
│   │   └── bootloader.bin
│   ├── config
│   │   └── sdkconfig.json
│   ├── flasher_args.json
│   ├── hello_world.bin
│   ├── hello_world.elf
│   ├── hello_world.map
│   └── partition_table
│       └── partition-table.bin
└── gdbinit
```

.github/workflows/python_tests.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install dependencies
         run: |
-          apt update && apt install -y socat
+          apt update && apt install -y socat && apt-get install libffi6
```
hfudev, line 28: the 38.0.0 wheel is available on https://dl.espressif.com/pypi/cryptography/

you could remove this lib header.

hfudev (changes requested):

pytest-embedded-idf/tests/test_idf.py
```diff
@@ -348,6 +348,46 @@ def test_detect_port(dut):
...
+    result.assert_outcomes(passed=1)
+
+
+def test_skip_check_panic_output(testdir):
+    testdir.makepyfile(r"""
+        import pexpect
+        import pytest
+
+        def test_check_panic_output(dut):
+            assert dut.expect('register dump')
```
hfudev, line 377: assert not here?

this is the same statement as the `test_check_panic_output` one `dut.expect('register dump')`. These two statement should be different, right?

pytest-embedded-idf/tests/test_idf.py
```diff
@@ -348,6 +348,46 @@ def test_detect_port(dut):
...
+
+    result.assert_outcomes(passed=1)
+
+
+def test_skip_check_panic_output(testdir):
+    testdir.makepyfile(r"""
+        import pexpect
+        import pytest
+
+        def test_check_panic_output(dut):
```
hfudev, line 376: wrong name here?

hfudev (changes requested):
@alekseiapa Thanks! This looks pretty nice. a few notes about the naming stuff. Let's avoid cli option breaking changes.

alekseiapa:
Hello @hfudev . Could you please take a look one more time ? I think it should be ready now

pytest-embedded-idf/tests/test_idf.py
```diff
@@ -4,6 +4,11 @@
 import pytest
 from pytest_embedded_idf.dut import IdfDut
 
+toolchain_required = pytest.mark.skipif(
+    os.getenv('PATH') is None or 'riscv32-esp-elf-gdb/bin' not in os.getenv('PATH'),
```
hfudev, line 8: the `'riscv32-esp-elf-gdb/bin'` won't work on windows since they're using a different dir separator. `os.path.join` instead

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -76,11 +91,73 @@ def toolchain_prefix(self) -> str:
...
+            os.getenv('IDF_PATH', 'IDF_PATH'),
+            'tools',
+            'gdb_panic_server.py',
+        )
+        if not os.path.isfile(script_filepath):
+            logging.info(
+                'Panic output decode script not found. Please use --panic-output-decode-script flag '
+                'to provide script or set IDF_PATH (Default: $IDF_PATH/tools/gdb_panic_server.py)'
+            )
+            raise ValueError
```
hfudev, line 110: raise ValueError('...') with your logging message above would be better.

The log message won't be printed if `--log-cli-level` is unset.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -76,11 +91,73 @@ def toolchain_prefix(self) -> str:
...
+        return os.path.realpath(script_filepath)
+
+    def _check_panic_decode_trigger(self):  # type: () -> None
+        with open(self.pexpect_proc._fr.name, 'rb') as output_file:
+            output = output_file.read()
+        panic_output_res = self.PANIC_OUTPUT_REGEX.search(output)
```
hfudev, line 116: large regex string often would cost a longer time for search. imaging such a use case.

A test script is expecting with a timeout as 30 seconds. The app run into core dump, which would make the app restart over and over again, for 30 seconds. The output of the file would be extremely large.

get the first "start line" and the first "end line", parse the string between them would be better.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -76,11 +91,73 @@ def toolchain_prefix(self) -> str:
...
+        if not os.path.isfile(script_filepath):
+            logging.info(
+                'Panic output decode script not found. Please use --panic-output-decode-script flag '
+                'to provide script or set IDF_PATH (Default: $IDF_PATH/tools/gdb_panic_server.py)'
+            )
+            raise ValueError
+        return os.path.realpath(script_filepath)
+
+    def _check_panic_decode_trigger(self):  # type: () -> None
+        with open(self.pexpect_proc._fr.name, 'rb') as output_file:
```
hfudev, line 114:
```suggestion
        with open(self.logfile, 'rb') as output_file:
```
don't access members startswith `_` unless you have to

pytest-embedded-idf/tests/test_idf.py
```diff
@@ -348,6 +353,60 @@ def test_detect_port(dut):
...
+            dut.expect(pexpect.TIMEOUT, timeout=5)
+    """)
+
+    result = testdir.runpytest(
+        '-s',
+        '--embedded-services', 'esp,idf',
+        '--app-path', f'{os.path.join(testdir.tmpdir, "hello_world_esp32c3_panic")}',
+        '--panic-output-decode-script', os.path.join(testdir.tmpdir, 'gdb_panic_server.py'),
+        '--skip-check-coredump', 'True',
+        '--target', 'esp32c3',
```
hfudev, line 398: unneeded.

pytest-embedded-jtag/tests/test_jtag.py
```diff
@@ -17,7 +17,7 @@ def test_pexpect_by_jtag(testdir):
         def test_pexpect_by_jtag(dut):
             dut.gdb.gdb_set('mi-async', 'on')
             dut.gdb.file_exec_and_symbols(os.path.join(dut.app.app_path, 'build', 'hello-world.elf'))
-            dut.gdb.interpreter_exec_console(f'source {os.path.join(dut.app.app_path, "gdbinit")}')
+            dut.gdb.interpreter_exec_console(f'source {os.path.join(dut.app.app_path, "prefix_map_gdbinit")}')
```
hfudev, line 20: gdbinit is the init commands to interact with openocd server

prefix_map_gdbinit is used when enable reproducible build.

hfudev (changes requested):
@alekseiapa Thanks for your contribution. LGTM! Left some minor comments

pytest-embedded-jtag/tests/test_jtag.py
alekseiapa: In this case  `IdfDut` instance doesn't have an `gdb` attribute. Should we add one more attribute to `IdfDut` ?

hfudev: here it's using the `JtagDut` as the dut class

alekseiapa: Oh. Sorry. My bad, seems I've changed this line by mistake without noticing it!

alekseiapa:
@hfudev Thank you for your review. All of the suggestions have been implemented.

hfudev:
Thanks!

Issue #127, opened 2022-08-26
use sphinx_idf_theme theme as the documentation template (RDT-268)


igrr:
To make pytest-embedded docs look similar to the docs of other projects at docs.espressif.com, we should consider using https://github.com/espressif/sphinx_idf_theme.

PR #128, opened 2022-09-01
docs: use readthedocs theme

hfudev:
related #127 

sphinx theme is highly binded to sphinx project. can't use that theme directly.

After changing the theme from `material` to `readthedocs` the layout turn out to be:

image


igrr (approved):
Looks like a good improvement! Maybe we need to make a version of the IDF team for mkdocs as well.
cc @ESP-Marius

PR #132, opened 2022-09-07
feat: use multiprocessing.Process with Queue

hfudev:


hfudev:
this one would work on windows for this prototype code

```python
# test.py
def test_dut(dut):
    for i in range(10):
        dut.expect('1')
```

```python
# conftest.py
import errno
import os
from pathlib import Path
from typing import BinaryIO

import pexpect.fdpexpect
import pytest
import multiprocessing
import time

from pexpect import TIMEOUT, EOF
from pexpect.utils import poll_ignore_interrupts, select_ignore_interrupts


@pytest.fixture
def filepath():
    return os.path.join(os.path.dirname(__file__), 'temp.txt')


@pytest.fixture
def msg_queue() -> multiprocessing.Queue:
    ctx = multiprocessing.get_context()

    return ctx.Queue()


def _listener(q: multiprocessing.Queue, f: str) -> None:
    while True:
        msg = q.get()
        with open(f, 'ab') as fw:
            fw.write(msg)
            fw.flush()
        time.sleep(0.1)


@pytest.fixture
def listener(msg_queue, filepath) -> multiprocessing.Process:
    os.makedirs(os.path.dirname(filepath), exist_ok=True)

    ctx = multiprocessing.get_context()

    return ctx.Process(
        target=_listener,
        args=(
            msg_queue,
            filepath,
        ),
        daemon=True,
    )


@pytest.fixture
def _pexpect_fr(filepath, listener) -> BinaryIO:
    Path(filepath).touch()
    listener.start()
    while True:
        if os.path.isfile(filepath):
            break
        else:
            time.sleep(0.1)
    return open(filepath, 'rb')


@pytest.fixture
def pexpect_proc(_pexpect_fr):
    return PexpectProcess(_pexpect_fr)


@pytest.fixture
def serial(msg_queue):
    return Serial(msg_queue)


@pytest.fixture
def dut(pexpect_proc, serial, msg_queue):
    return Dut(pexpect_proc, msg_queue)


class PexpectProcess(pexpect.fdpexpect.fdspawn):
    def __init__(self, pexpect_fr):
        super().__init__(pexpect_fr)

    def read_nonblocking(self, size=1, timeout=-1) -> bytes:
        """
        Since we're using real file stream, here we only raise an EOF error only when the file stream has been closed.
        This could solve the `os.read()` blocked issue.

        Args:
            size: most read bytes
            timeout: timeout

        Returns:
            read bytes
        """
        try:
            if os.name == 'posix':
                if timeout == -1:
                    timeout = self.timeout
                rlist = [self.child_fd]
                wlist = []
                xlist = []
                if self.use_poll:
                    rlist = poll_ignore_interrupts(rlist, timeout)
                else:
                    rlist, wlist, xlist = select_ignore_interrupts(rlist, wlist, xlist, timeout)
                if self.child_fd not in rlist:
                    raise TIMEOUT('Timeout exceeded.')

            s = os.read(self.child_fd, size)
        except OSError as err:
            if err.args[0] == errno.EIO:  # Linux-style EOF
                pass
            if err.args[0] == errno.EBADF:  # Bad file descriptor
                raise EOF('Bad File Descriptor')
            raise

        s = self._decoder.decode(s, final=False)
        self._log(s, 'read')
        return s


class Dut:
    def __init__(self, pexpect_proc, msg_queue):
        self.p = pexpect_proc
        self.q = msg_queue

    def expect(self, *args, **kwargs):
        self.p.expect(*args, **kwargs)

    def write(self, s):
        self.q.put(s)


class Serial:
    def __init__(self, msg_queue):
        self.q = msg_queue

        ctx = multiprocessing.get_context()
        p = ctx.Process(target=self.redirect)
        p.start()

    def redirect(self):
        while True:
            self.q.put(b'1')
            time.sleep(0.1)
```

hfudev:
oh, it won't work if reading from the serial.

the pyserial object can't be pickled.

hfudev:
```python
class Serial(multiprocessing.Process):
    def __init__(self, msg_queue):
        self.proc = None
        self.q = msg_queue
 
        super().__init__(target=self.redirect, args=('COM3',))
 
    def redirect(self, port):
        self.proc = pyserial.serial_for_url(port)
        while self.proc.is_open:
            s = self.proc.read_all()
            logging.debug(s)
            self.q.put(s)
            time.sleep(0.1)
```

This would work :D

PR #133, opened 2022-09-09
Refactor/multiprocessing queue and serial process

### LINKED ISSUE #67: Get Rid of Thread (RDT-106)
Thanks to GIL, using threads to redirect stdout internally could cause performance loss if the test script is also using `Thread`.

Try to find a way to stop using threads.

hfudev:
closes #67


hfudev (commented):
ci: change wheel

Issue #105, opened 2022-07-14
[idf,jtag,qemu] Support running tests built as loadable ELF files (RDT-236)


igrr:
Some test applications are small enough to be loaded into RAM of the target chip, and don't have to be flashed. Loading into RAM is typically faster than flashing.

The task is to support running tests with loadable ELF files for `idf`, `jtag` and `qemu` plugins.

For the ESP chips, ELF files can be loaded into the chip RAM in two ways:

- Using `esptool.py`. In this case, the ELF file first needs to be converted into a .bin file (using esptool.py elf2image). Then the resulting .bin file can be loaded into chip RAM using `esptool.py load_ram`. The program is loaded and immediately starts to execute.
- Over JTAG. In this case, the program can be loaded into RAM using GDB `load` command. (This approach works for all chips which have OpenOCD support, not necessarily ESPs.)

Note 1: we need to be careful not to reset the chip after loading the program and before running the test. Unlike the case when the program is flashed to the chip, the program loaded into RAM will be lost after reset. Then the chip will execute whichever program is stored in flash (if any).

Note 2: not every ELF file can be loaded into ESP chip RAM using the options above. When building an app using ESP-IDF, the option `CONFIG_APP_BUILD_TYPE_ELF_RAM` should be enabled. If it is not enabled, the resulting ELF file can't be loaded into RAM. 
It is also possible to build loadable ELF files without IDF using the latest esp-2022r1-rc1 GCC based toolchain. (For the ESP32 only, so far.) For instructions, see internal merge request idf/newlib-cygwin!44.

For the old TTFW based implementation, see https://github.com/espressif/esp-idf/blob/master/tools/test_apps/system/gdb_loadable_elf/app_test.py.

PR #140, opened 2022-09-23
Refactor/jtag

hfudev:
This is a pre-requisite MR for #105


pytest-embedded-jtag/pytest_embedded_jtag/openocd.py
```diff
@@ -1,36 +1,65 @@
...
+                f'gdb_port {self.gdb_port}',
+            ]
+        )
+        if app and hasattr(app, 'flash_files'):  # IdfApp
+            # flash via jtag
+            for _f in app.flash_files:
+                if _f.encrypted:
+                    raise ValueError('Encrypted files can\'t be flashed in via JTAG')
+
+                openocd_cli_args.extend(['-c', f'program_esp {_f.file_path} {hex(_f.offset)} verify'])
```
igrr, line 60: If there are multiple tests, will this happen on every test or just once per test app?

pytest-embedded-jtag/pytest_embedded_jtag/gdb.py
```diff
@@ -1,54 +1,21 @@
+import logging
+import shlex
 from typing import Optional
 
 from pytest_embedded.log import DuplicateStdoutPopen, MessageQueue
 
 
 class Gdb(DuplicateStdoutPopen):
     GDB_PROG_PATH = 'xtensa-esp32-elf-gdb'
-    GDB_DEFAULT_ARGS = '--nx --quiet --interpreter=mi2'
```
igrr, line 8: Will switching mi2 mode off make it harder to get information (like backtraces) in a structured way?

pytest-embedded-jtag/pytest_embedded_jtag/openocd.py
hfudev: cache added. thanks!

pytest-embedded-jtag/pytest_embedded_jtag/gdb.py
hfudev: mi2 would print a json-like bytestring. since we're using `expect` functions everywhere, normal string output would be more user-friendly IMHO...

if the user need to do it in the mi2 way, could override via CLI option `--gdb-cli-args [ARG]`

igrr: I think `expect` is fine for unstructured text output (like UART logs), but if we can get the information (like a backtrace) in a structured way via pygdbmi (like we do in IDF panic tests, for instance), I think it would generally be more reliable.

So my suggestion is rather to use mi2 together with pygdbmi, rather than mi2 with expect-like matching.

igrr: Actually these could be two different ways we set up the DUT, one could use text-mode GDB interactions, the other could use MI2 and pygdbmi.

hfudev: created #143 to do this in a separate PR.

Issue #117, opened 2022-08-01
[esp] Support bind Serial Port by `read_mac` (RDT-257)


hfudev:
esp devices support `esptools -p [PORT] read_mac` to get the MAC address.

Use this to specify the port.

hfudev:
consider PR to https://github.com/pyserial/pyserial

PR #141, opened 2022-09-29
feat: Added support for binding ports based on the USB device location

alekseiapa:
- Added support for binding ports based on the location of a USB device

Closes https://github.com/espressif/pytest-embedded/issues/118

PR #145, opened 2022-10-11
feat: Added support for binding ports based on the MAC address of the board

alekseiapa:
- Added support for binding ports based on the MAC address of the board

Closes https://github.com/espressif/pytest-embedded/issues/117

### DIFF
--- a/pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
+++ b/pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
@@ -1,6 +1,7 @@
 import contextlib
 import functools
 import logging
+import subprocess
 from typing import Optional
 
 import esptool
@@ -9,9 +10,24 @@ from esptool import CHIP_DEFS, FatalError
 from esptool import __version__ as ESPTOOL_VERSION
 from esptool import detect_chip
 from esptool.targets import CHIP_LIST as ESPTOOL_CHIPS
-from pytest_embedded.log import MessageQueue
+from pexpect import TIMEOUT
+from pytest_embedded.log import MessageQueue, _PexpectProcess, live_print_call
 from pytest_embedded.utils import Meta
 from pytest_embedded_serial.dut import Serial
+
+
+def _is_port_mac_verified(pexpect_proc: _PexpectProcess, port: str, port_mac: str, msg_queue) -> bool:
+    try:
+        live_print_call(['esptool.py', '--port', port, 'read_mac'], msg_queue=msg_queue)
+    except subprocess.CalledProcessError:
+        return False
+    else:
+        try:
+            pexpect_proc.expect(f'MAC: {port_mac.lower()}', timeout=0.1)
+        except TIMEOUT:
+            return False
+        else:
+            return True
 
 
 class EsptoolArgs(object):
@@ -33,10 +49,12 @@ class EspSerial(Serial):
 
     def __init__(
         self,
+        pexpect_proc: _PexpectProcess,
         msg_queue: MessageQueue,
         target: Optional[str] = None,
         beta_target: Optional[str] = None,
         port: Optional[str] = None,
+        port_mac: str = None,
         baud: int = Serial.DEFAULT_BAUDRATE,
         esptool_baud: int = ESPTOOL_DEFAULT_BAUDRATE,
         skip_autoflash: bool = False,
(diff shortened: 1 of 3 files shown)


alekseiapa:
Hello @hfudev . Could you please take a look at the current PR ? Thanks!

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -54,6 +67,13 @@ def __init__(
             # sort to make /dev/ttyS* ports before /dev/ttyUSB* ports
             # esptool will reverse the list
             ports.sort()
+            if port_mac:
+                for port in available_ports:
```
hfudev, line 71: here's a potential bug. in earlier lines,

```python
            available_ports = esptool.get_port_list()
            ports = list(set(available_ports) - set(self.occupied_ports.keys()))
```

should use `ports` instead. the occupied ports should not be considered.

This potential bug is also happening in PR #141, please also fix it there. sorry I missed that in earlier review.

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -14,6 +15,17 @@
...
 
+def is_port_mac_verified(port: str, port_mac: str) -> bool:
+    try:
+        output = subprocess.check_output(['esptool.py', '--port', port, 'read_mac'], stderr=subprocess.STDOUT).decode()
+    except subprocess.CalledProcessError:
+        return False
+    if f'MAC: {port_mac.lower()}' in output:
+        return True
+    else:
+        return False
```
hfudev, line 26: please use `live_print_call(cmd, msg_queue=msg_queue)` and `dut.expect(f'MAC: {port_mac.lower()}')` inside the `EspSerial` class

The benefit to use `live_print_call` is that:
1. it would print the stdout to the terminal (most important)
2. it would check the return code automatically
3. you could do some expect on the output

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -65,7 +85,15 @@ def __init__(
 
             logging.debug(f'Detecting ports from {", ".join(ports)}')
         else:
-            ports = [port]
+            if port_mac:
+                if is_port_mac_verified(port, port_mac):
+                    ports = [port]
+                else:
+                    raise ValueError(
+                        f'The specified MAC address {port_mac} binds with different port, ' f'not with {port}'
```
hfudev, line 93: format a bit weird?
```suggestion
                        f'The specified MAC address {port_mac} binds with different port, not with {port}'
```

hfudev (changes requested):
@alekseiapa overall LGTM! Thanks!

hfudev (commented):
thanks! LGTM. only some minor issues.

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -9,11 +10,26 @@
...
 from esptool import detect_chip
 from esptool.targets import CHIP_LIST as ESPTOOL_CHIPS
-from pytest_embedded.log import MessageQueue
+from pexpect import TIMEOUT
+from pytest_embedded.log import MessageQueue, _PexpectProcess, live_print_call
 from pytest_embedded.utils import Meta
 from pytest_embedded_serial.dut import Serial
 
 
+def is_port_mac_verified(pexpect_proc: _PexpectProcess, port: str, port_mac: str, msg_queue) -> bool:
```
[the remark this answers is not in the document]
alekseiapa: Then I have to create the instance attributes for all of these parameters since at the moment when this function called there is no any of these attributes available.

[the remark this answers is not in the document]
hfudev: I agree. how about making this function starts with `_`? in order to not be generated in the API reference.

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -997,6 +1011,7 @@ def _fixture_classes_and_options(
 
                 classes[fixture] = Serial
                 kwargs[fixture] = {
+                    'pexpect_proc': pexpect_proc,
```
hfudev, line 1014: This line should be removed. Only `EspSerial` is using this param.

hfudev (commented):
LGTM.

besides, your commit message should follow https://www.conventionalcommits.org/en/v1.0.0/ somehow. 

- fix(serial): remove any occupied ports from ...
- feat(esp): support bind ports by esptool `read_mac`

Starting with a verb would be better.

alekseiapa:
Can't pass the test `test_expect_from_timeout` for both python environments. But it is not related the code change

hfudev:
Thanks! @alekseiapa

Issue #142, opened 2022-10-06
Record Unity test execution time into JUnit report (RDT-289)


igrr:
Currently add_unity_test_cases function parses the complete output of all the test cases and produces the Junit report. One thing we aren't doing with this approach is getting the execution times of individual test cases. This can be helpful when analyzing CI bottlenecks, to find specific test cases which are running for a long time.

PR #153, opened 2022-10-27
feature: move test case runner. record unity test execution time into JUnit report

alekseiapa:
- Added unity test execution time into JUnit report
- Moved test_case_runner
- Added tests for test_case_runner


hfudev:
@alekseiapa https://stackoverflow.com/questions/37471740/how-to-copy-commits-from-one-git-repo-to-another

alekseiapa:
> @alekseiapa https://stackoverflow.com/questions/37471740/how-to-copy-commits-from-one-git-repo-to-another

@hfudev Yes, this is the way how I've done that. The author is @L-KAYA and the committer is me since I've changed the commit message to make it more appropriate for the current repository

hfudev:
> > @alekseiapa https://stackoverflow.com/questions/37471740/how-to-copy-commits-from-one-git-repo-to-another
> 
> @hfudev Yes, this is the way how I've done that. The author is @L-KAYA and the committer is me since I've applied the commit to the current history and changed the commit message to make it more appropriate for the current repository. The commit history is totally preserved here

sorry! please ignore!

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -86,19 +100,20 @@ def run_normal_case(self, case: UnittestMenuCase, reset: bool = False, timeout:
...
                 self.dut.write(str(case.index))
                 try:
-                    self.dut.expect('Running {}...'.format(case.name), timeout=1)
+                    self.dut.expect(f'Running {case.name}...', timeout=1)
                     break
                 except TIMEOUT as e:
-                    if retry >= self.retry_times - 1:
+                    if retry >= retry_count - 1:
                         raise e
             self.dut.expect_unity_test_output(timeout=timeout)
```
hfudev, line 118: should also retry if this test case failed. here's the logic is a bit more complicated.

- if the result is passed:
    return success
- if the result is failed (or no result since core dumped or whatever) but retried:
    not decided yet
- if the result is failed (or no result since core dumped or whatever) and the retry_count has exceeded:
   return failed

hfudev: this try...except... block should wrap the whole block, like

```python
try:
    self.dut.write(str(case.index))
    self.dut.expect(f'Running {case.name}...', timeout=1)
    self.dut.expect_unity_test_output(timeout=timeout)
except ...:
    ...
```

besides, the timeout check should also calculate the whole process., use some techniques like https://docs.python.org/3/library/threading.html#threading.Timer

hfudev: same logic to other `run_...` functions

hfudev (changes requested):
@alekseiapa Left a few comments about the retry logic.

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,443 @@
...
+        dut.write(str(sub_case_index))
+
+        # Wait for the specific patterns, only exist when the sub-case finished
+        while True:
+            pat = dut.expect(signal_pattern_list, timeout=60)
+            if pat is not None:
+                match_str = pat.group().decode('utf-8')
+
+                # Send a signal
+                if match_str.find(self.SEND_SIGNAL_PREFIX) >= 0:
```
L-KAYA, line 316: As a noob, I learnt that this might be simplified to `if self.SEND_SIGNAL_PREFIX in match_str:`. Same as following `elif`

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,443 @@
...
+
+
+class BaseTester:
+    """
+    The base class that providing shared methods
+
+    Attributes:
+        dut (Dut): Object of the Device under test
+        test_menu (List[UnittestMenuCase]): The list of the cases
+        args (Any): Not used
```
L-KAYA, line 20: I forgot to delete this line, args is deleted

L-KAYA (commented):
Thanks for moving this script into pytest-embeded!

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
alekseiapa: I suppose this logic can be applied to the `run_normal_case` only. For `run_multistage` and `run_multidev` it requires to pass the `sub_case_index`, so the `dut.write(str(sub_case_index))` can't be placed under `try` block

hfudev: what do you mean by "can't be placed under the try block" since those two functions "requires to pass the sub_case_index"?

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,452 @@
...
+        """
+        if case.type == 'normal':
+            if reset:
+                self.dut.serial.hard_reset()
+            self.dut.expect(self.ready_pattern_list, timeout=timeout)
+            for retry in range(retry_count):
+                try:
+                    self.dut.write(str(case.index))
+                    self.dut.expect(f'Running {case.name}...', timeout=1)
+                    self.dut.expect_unity_test_output(timeout=timeout)
```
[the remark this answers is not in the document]
alekseiapa: @igrr Is your idea to only apply the retry logic at the start of the test ? But when we are checking the unity test output we shouldn't retry . Is that what you meant ?

alekseiapa:
Hello @hfudev . Could you please take a look at the changes ? Thanks!

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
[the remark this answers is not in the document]
L-KAYA: When I tested the script in IDF, I found writing case index to dut might fail  after reset. I guess the failure mainly caused by the hurry writting before the dut get totally ready to receive something, because some delays will help. However, I can't set an exact delay since the required delay are different among different environment, especially on CI.  Therefore, I retired when failed to expect case running pattern.

As for the `expect_unity_test_output` I agree that it shouldn't be included in the `retry` block, because after expecting the `case running` pattern, the case is supposed to be running, when the timeout happened then, it should be treated as a failure.
cc @igrr @hfudev

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
L-KAYA: > 



> this try...except... block should wrap the whole block, like
> 
> ```python
> try:
>     self.dut.write(str(case.index))
>     self.dut.expect(f'Running {case.name}...', timeout=1)
>     self.dut.expect_unity_test_output(timeout=timeout)
> except ...:
>     ...
> ```
> 
> besides, the timeout check should also calculate the spent time of the code block above. You may use https://docs.python.org/3/library/threading.html#threading.Timer or anything similar.

I would suggest to remain the legacy logic. Reasons given in https://github.com/espressif/pytest-embedded/pull/153#discussion_r1009061388

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,452 @@
...
+
+
+def timeit(func):
+    @wraps(func)
+    def timeit_wrapper(*args, **kwargs):
+        start_time = time.perf_counter()
+        result = func(*args, **kwargs)
+        end_time = time.perf_counter()
+        total_time = end_time - start_time
+        logging.info(f'Testcase "{args[1].name}" took {total_time:.5f} seconds')
```
hfudev, line 22: logging doesn't help. should raise a timeout and terminate running the current test case.

hfudev (changes requested):

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
alekseiapa: I agree with Raising timeout. My plan was to use this line later to track test execution time in the junit report.

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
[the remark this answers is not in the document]
hfudev: the original idea of the `retry` is for those flaky unit test cases. it should be parsed from the test case's full name. like `TEST_CASE(...[retry=3][timeout=120])`

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
hfudev: this timeout would also help kill those hanging processes. that's why raise immediately is better, IMHO

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
hfudev: I think we're talking about two things here... Let me clarify it a bit:

- for the idea to support keywords defined with `TEST_CASE` (described here: https://github.com/espressif/pytest-embedded/pull/153#discussion_r1009090953), please add the retry mechanism to the whole block
- for the idea to run retry while starting the test case, I agree with @L-KAYA 's idea. We could improve it later in a separate PR.

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -1,13 +1,29 @@
...
 
+DEFAULT_RETRY_COUNT = 3
+DEFAULT_TIMEOUT = 90
+
+
+def timeit(func):
+    @wraps(func)
+    def timeit_wrapper(*args, **kwargs):
+        start_time = time.perf_counter()
+        result = func(*args, **kwargs)
```
alekseiapa, line 19: @hfudev  How about I apply the retry logic for the case in decorator: 
```python
def timeit(func):
    @wraps(func)
    def timeit_wrapper(*args, **kwargs):
        retry_count = int(kwargs.get('timeout', DEFAULT_RETRY_COUNT))
        start_time = time.perf_counter()
        for retry in range(retry_count):
            try:
                result = func(*args, **kwargs)
                break
            except TIMEOUT as e:
                if retry >= retry_count - 1:
                    raise e
        end_time = time.perf_counter()
        total_time = end_time - start_time
        logging.info(f'Testcase "{args[1].name}" took {total_time:.5f} seconds')

        if total_time > int(kwargs.get('timeout', DEFAULT_TIMEOUT)):
            raise TIMEOUT

        return result

    return timeit_wrapper
```

alekseiapa:
@hfudev Could you please take a look at the changes ? Thank you!

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -1,13 +1,40 @@
...
+                break
+            except TIMEOUT as e:
+                if retry >= retry_count - 1:
+                    raise e
+        end_time = time.perf_counter()
+        total_time = end_time - start_time
+        logging.info(f'Testcase "{args[1].name}" took {total_time:.5f} seconds')
+
+        if total_time > int(kwargs.get('timeout', DEFAULT_TIMEOUT)):
+            raise TIMEOUT
```
hfudev, line 32: > this timeout would also help kill those hanging processes. that's why raise immediately is better, IMHO

as I mentioned above, this current method can't kill the hanging processes.

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -1,13 +1,40 @@
...
 from pytest_embedded_idf.dut import UnittestMenuCase
 
+DEFAULT_RETRY_COUNT = 3
+DEFAULT_TIMEOUT = 90
+
+
+def timeit(func):
+    @wraps(func)
+    def timeit_wrapper(*args, **kwargs):
+        retry_count = int(kwargs.get('timeout', DEFAULT_RETRY_COUNT))
```
hfudev, line 18: - The `retry_count` here should be defined at `UnittestMenuCase.attributes['retry']`
- timeout should be defined at `UnittestMenuCase.attributes['timeout']`

> the original idea of the retry is for those flaky unit test cases. it should be parsed from the test case's full name. like TEST_CASE(...[retry=3][timeout=120])

As I mentioned above, these keywords should be defined in the `TEST_CASE(...)` c file. Not a keyword for python function.

hfudev (changes requested):

alekseiapa:
@hfudev I've added the logic of retrying the test case to the decorator. The options`UnittestMenuCase.attributes['retry']` and `UnittestMenuCase.attributes['timeout']` are parsed in `run_test` method

alekseiapa:
@hfudev I've decided to resolve issue [#142](https://github.com/espressif/pytest-embedded/issues/142) in the current PR because it is related to the current task. What are your thoughts about it?

hfudev:
Overall LGTM. btw, you don't need to mention the internal tickets here. I've removed them.

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,448 @@
...
+        """
+        Get the test menu of this test app
+
+        Notes:
+            It will do a hard reset after getting the test menu to ensure
+            the patterns that indicate the case is ready not taken by the parser.
+            Please use this function to get the test menu while using this script
+        """
+        self.dut.write('')
+        self.test_menu = self.dut.parse_test_menu()
```
hfudev, line 52: Instance attribute test_menu defined outside `__init__ `

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,448 @@
...
+        self,
+        reset: bool = False,
+        timeout: int = DEFAULT_TIMEOUT,
+        retry_count: int = DEFAULT_RETRY_COUNT,
+    ) -> None:
+        """
+        Run all normal cases
+
+        Args:
+            reset: whether do a hardware reset before running the case
```
hfudev, line 76:
```suggestion
            reset: whether to do a hardware reset or not before running the case
```

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,448 @@
+import time
```
hfudev, line 1: rename this file to `unity_tester.py`, two `idf` is a bit redundant

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,448 @@
...
+    """
+
+    def __init__(self, dut: Dut) -> None:
+        self.dut = dut
+        self.sem = Semaphore()
+        self.recv_sig: List[str] = []
+        self.thread: Thread = None  # type: ignore
+
+
+class MultiDevCaseTester(BaseTester):
```
hfudev, line 203: could you help create a GH issue "replace threads in `MultiDevCaseTester`" later after your PR got merged?

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,448 @@
...
+
+    def __init__(self, dut: Union[Dut, List[Dut]], **kwargs) -> None:  # type: ignore
+        """
+        Create the object for every dut and put them into the group
+        """
+        super().__init__(dut, **kwargs)
+        self.group: List[MultiDevResource] = []
+        if isinstance(dut, List):
+            for item in dut:
+                if isinstance(item, Dut):
```
hfudev, line 228: this if clause seems a bit too defensive... consider remove it.

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,448 @@
...
+    """
+
+    # The patterns that indicate the runner is ready come from 'unity_runner.c'
+    ready_pattern_list = [
+        'Press ENTER to see the list of tests',
+        'Enter test for running',
+        'Enter next test, or \'enter\' to see menu',
+    ]
+
+    def __init__(self, dut: Union[Dut, List[Dut]], **kwargs) -> None:  # type: ignore
```
hfudev, line 29:
```suggestion
    def __init__(self, dut: IdfDut, **kwargs) -> None:  # type: ignore
```

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
```diff
@@ -0,0 +1,448 @@
...
+    Resources of multi_dev dut
+
+    Attributes:
+        dut (Dut): Object of the Device under test
+        sem (Semaphore): Semaphore of monitoring whether the case finished
+        recv_sig (List[str]): The list of received signals from other dut
+        thread (Thread): The thread of monitoring the signals
+    """
+
+    def __init__(self, dut: Dut) -> None:
```
hfudev, line 196:
```suggestion
    def __init__(self, dut: IdfDut) -> None:
```

hfudev (changes requested):
a few more minor comments.

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
alekseiapa: I guess here it still has to be Union: 
`def __init__(self, dut: Union[IdfDut, List[IdfDut]], **kwargs) -> None:  # type: ignore`

hfudev: well, I think only `MultiDevTester` requires multiple duts?

let's keep it as what it is right now, and improve later.

pytest-embedded-idf/pytest_embedded_idf/idf_unity_tester.py
hfudev: >  after your PR got merged

anyway... I asked so because if this PR is not merged, the issue itself is meaningless.

alekseiapa:
@hfudev Could you please take a look ? The pipeline is passing now

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -6,37 +6,37 @@
...
-                          'Enter next test, or \'enter\' to see menu']
+    ready_pattern_list = [
+        'Press ENTER to see the list of tests',
+        'Enter test for running',
+        'Enter next test, or \'enter\' to see menu',
+    ]
 
-    def __init__(self, dut: Union[Dut, List[Dut]], **kwargs) -> None:     # type: ignore
-        self.retry_times = 30
+    def __init__(self, dut: Union[Dut, List[Dut]], **kwargs) -> None:  # type: ignore
```
hfudev, line 31: IdfDut

hfudev (approved):

hfudev:
thank you! @alekseiapa

PR #157, opened 2022-11-04
fix: Added test case name uniqueness validation

alekseiapa:
Added test case name uniqueness validation

### DIFF
--- a/pytest-embedded/pytest_embedded/plugin.py
+++ b/pytest-embedded/pytest_embedded/plugin.py
@@ -1347,6 +1347,15 @@ class PytestEmbedded:
 
     @pytest.hookimpl(trylast=True)
     def pytest_collection_modifyitems(self, items: List[Function]):
+        def _check_duplicate_items(names_list: List[str]):
+            if len(names_list) != len(set(names_list)):
+                duplicates = set([name for name in names_list if names_list.count(name) > 1])
+                if duplicates:
+                    raise ValueError(f'Duplicated items: {duplicates}')
+
+        _check_duplicate_items([test.name for test in items])
+        _check_duplicate_items([os.path.basename(name) for name in set([str(test.path.absolute()) for test in items])])
+
         if self.parallel_index == 1 and self.parallel_count == 1:
             return
 
--- a/pytest-embedded/tests/test_base.py
+++ b/pytest-embedded/tests/test_base.py
@@ -1,5 +1,6 @@
 import os
 import xml.etree.ElementTree as ET
+from dataclasses import dataclass
 from pathlib import Path
 
 import pytest
@@ -165,7 +166,14 @@ def test_default_app_path(testdir):
 def test_parallel_run(parallel_count, parallel_index, res):
     from pytest_embedded.plugin import PytestEmbedded
 
-    fake_items = [1, 2, 3, 4, 5]
+    @dataclass
+    class _FakeItem:
+        name: str
+        path = Path('.')
+
+    # _Fake instances are used in order to pass `pytest_collection_modifyitems` hook
+    fake_items = [_FakeItem(name='1'), _FakeItem(name='2'), _FakeItem(name='3'),
+                  _FakeItem(name='4'), _FakeItem(name='5')]
     fake_plugin = PytestEmbedded(parallel_count, parallel_index)
     fake_plugin.pytest_collection_modifyitems(fake_items)  # noqa
     assert len(fake_items) == res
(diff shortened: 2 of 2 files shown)


hfudev:
@alekseiapa no need to do add the internal task id here. I've removed that.

alekseiapa:
@hfudev could you please take a look at the MR ? The workflow failure is not related to the code change here.

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1335,6 +1335,23 @@ def pytest_fixture_setup(self, fixturedef: FixtureDef[Any], request: SubRequest)
...
+                    logging.error(
+                        f'A duplicate {query} name was discovered: {duplicate}. ' f'Each test {query} should be unique'
+                    )
+                exit(1)
+
+        case_name_list = [test.name for test in items]
+        case_module_list = [str(test.path.absolute()) for test in items]
+
+        _check_duplicate_name(case_name_list, query='case name')
+        _check_duplicate_name(case_module_list, query='module name')
```
hfudev, line 1353: Sorry to point it out, but this kind of abstraction is actually making the code harder to read... you can actually put

```python
if query == 'module name':
    path_set = set(names_list)
    names_list = [os.path.basename(name) for name in path_set]
```

into the lines that make the actual call.

pytest-embedded/tests/test_base.py
```diff
@@ -165,7 +166,14 @@ def test_default_app_path(app):
 def test_parallel_run(parallel_count, parallel_index, res):
     from pytest_embedded.plugin import PytestEmbedded
 
-    fake_items = [1, 2, 3, 4, 5]
+    @dataclass
+    class _FakeItem:
```
hfudev, line 170: or we could use the `_pytest.nodes.Item` instead.

hfudev: Otherwise, we can't guarantee there won't be any breaking changes some day...

hfudev (changes requested):
overall lgtm. could you please add a test for that?

`testdir.makepyfile` could help generate same module name files under different packages

and also a test to check the duplicated test function name.

pytest-embedded/pytest_embedded/plugin.py
alekseiapa: Here I am not able to do that since variable `query` and the value of this variable doesn't have any source. This var is created by me

hfudev:
```python
        def _check_duplicate_name(l: List[str]):
            if len(l) != len(set(l)):
                duplicates = set([name for name in l if l.count(name) > 1])
                if duplicates:
                    raise ValueError(f'Duplicated items: {duplicates}')
                exit(1)

        case_name_list = [test.name for test in items]
        case_module_list = [str(test.path.absolute()) for test in items]

        _check_duplicate_name(case_name_list)
        _check_duplicate_name([os.path.basename(name) for name in set(case_module_list)])
```
like this. Move the `query` logic into the calling statements, not the common function.

hfudev: even shorter, var `case_name_list` and `case_module_list` are not necessarily needed.

pytest-embedded/tests/test_base.py
alekseiapa: According to documentation `Direct construction of _pytest.nodes.Item has been deprecated, please use _pytest.nodes.Item.from_parent.` >> [doc](https://docs.pytest.org/en/stable/deprecations.html#node-construction-changed-to-node-from-parent) .

I think we can leave it as it is since the `_pytest.nodes.Item.from_parent()` approach much larger than the current one

hfudev: ah, good to know. I agree.

alekseiapa:
@hfudev The tests to check the function's functionality have been added. Would you kindly look at this?

hfudev (changes requested):
LGTM! only minor comments.

pytest-embedded/tests/test_base.py
```diff
@@ -467,3 +475,37 @@ def test_set_log_extension(dut):
...
+        """)
+    os.rename(f'{dup_module_path}', os.path.join(test_sub_dir, dup_module_path.basename))
+    testdir.makepyfile(test_duplicate_module=r"""
+                    def test_duplicate_two():
+                        pass
+
+                """)
+    testdir.runpytest()
+
+    assert "ValueError: Duplicated items: {'test_duplicate_module.py'}" in capsys.readouterr().out
```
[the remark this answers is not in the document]
alekseiapa: This way will not work since the error will be raised before the test execution will be started so the test case itself will not start properly

alekseiapa:
@hfudev The threads have been resolved except the one regarding Assertion check

pytest-embedded/tests/test_base.py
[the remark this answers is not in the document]
hfudev: get it. Thanks!

hfudev:
Thank you!

PR #166, opened 2022-11-25
feat: add `run_all_single_board_cases` to `IdfDut`

### LINKED ISSUE #142: Record Unity test execution time into JUnit report (RDT-289)

### LINKED ISSUE #162: [idf] replace threads in MultiDevCaseTester (RDT-317)
Currently we use threads for the serial output redirection purposes. Python Global Interpreter Lock might cause the block of the threads. In order to avoid block we should reduce their amount

hfudev:
- record execution time for failed/unfinished unity test cases (closes #142)


hfudev:
@L-KAYA PTAL as well

alekseiapa:
@hfudev I went through all of the changed lines and couldn't come up with any more questions. Thanks!

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -260,20 +249,43 @@ def parse_test_menu(
...
-        Parse test case mcnu from string to list of `UnittestMenuCase`.
+        Parse test case menu from string to list of `UnittestMenuCase`.
 
         Args:
             s: string include test case menu.
 
         Returns:
             A `list` of `UnittestMenuCase`, which includes info for each test case.
         """
-        print(s)
```
L-KAYA, line 276: Nice deletion!

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -335,6 +347,204 @@ def parse_unity_menu_from_str(s: str) -> List[UnittestMenuCase]:
 
         return test_menu
 
+    @property
+    def test_menu(self) -> t.List[UnittestMenuCase]:
+        if self._test_menu is None:
+            self._test_menu = self._parse_test_menu()
+            self.serial.hard_reset()
```
L-KAYA, line 354: Do you think it's necessary to log that the test menu is parsed successfully before restart? In case that the restart is caused by other reasons.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -335,6 +347,204 @@ def parse_unity_menu_from_str(s: str) -> List[UnittestMenuCase]:
...
+        self,
+        case: UnittestMenuCase,
+        reset: bool = False,
+        timeout: float = 30,
+    ) -> None:
+        """
+        Run a specific multi_stage case
+
+        Notes:
+            Will skip with a warning if the case type is not normal
```
L-KAYA, line 475:
```suggestion
            Will skip with a warning if the case type is not multi stage
```

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -335,6 +347,204 @@ def parse_unity_menu_from_str(s: str) -> List[UnittestMenuCase]:
...
+        Args:
+            case: the specific case that parsed in test menu
+            reset: whether to perform a hardware reset before running a case
+            timeout: timeout. (Default: 30 seconds)
+        """
+        if case.type != 'multi_stage':
+            logging.warning('case %s is not a multi stage case', case.name)
+            return
+
+        _start_at = time.perf_counter()
```
L-KAYA, line 486: Why we still need to record the timestamp? Doesn't it have been recorded in `_record_single_unity_test_case `?

L-KAYA (approved):
LGTM, thanks for optimizing the script!

pytest-embedded-idf/pytest_embedded_idf/dut.py
hfudev: to calculate the remain timeout.

hfudev:
@alekseiapa @L-KAYA Thank you all for your review!

PR #169, opened 2022-11-30
fix(idf): set correct toolchain prefix for RISC-V targets

igrr:
Previously dut.toolchain_prefix would return `riscv32-esp32c3-elf-`. However for RISC-V, the toolchain for all targets has a prefix `riscv32-esp-elf-`.

Also simplify the invocation of GDB since we can rely on `{toolchain_prefix}-gdb` program to be available for all targets.


hfudev:
Thanks! @igrr

PR #176, opened 2022-12-27
feature: prettify xml junit report

### LINKED ISSUE #165: Prettify the Generated Junit Report (RDT-324)

alekseiapa:
- prettify xml junit report
close #165

### DIFF
--- a/pytest-embedded/pytest_embedded/plugin.py
+++ b/pytest-embedded/pytest_embedded/plugin.py
@@ -9,6 +9,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import xml.dom.minidom
 from collections import Counter, defaultdict, namedtuple
 from operator import itemgetter
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


alekseiapa:
@hfudev Could you please take a look ?

hfudev:
LGTM. could you set it to a CLI option? and make it non default. the default one should still be the minified version.

btw, please also `close #165`

alekseiapa:
@alekseiapa Could you please take a look ? I'll close #165 once this PR got merged

hfudev:
You may add this inside the PR description to auto-close the related issue.

pytest-embedded/pytest_embedded/utils.py
```diff
@@ -80,6 +81,13 @@ def find_by_suffix(suffix: str, path: str) -> List[str]:
     return res
 
 
+def prettify_xml(file_path: str):
```
hfudev, line 84: this function will never be reused... how about we move it to the `plugin.py` instead?

hfudev (changes requested):
last comment. Thank you!

pytest-embedded/pytest_embedded/utils.py
alekseiapa: Sure, I've moved the function to the correct place.

PR #179, opened 2023-01-04
Refactor/make idf common functions as mixin

### LINKED ISSUE #164: [qemu,idf] `QemuDut` based on `Dut` even when service `idf` is enabled (RDT-319)
- put some IDF non-real-target related functions as a mixin
- if `idf` service is enabled, inject the mixin into the `QemuDut` class

hfudev:
- closes #164 
- created #205


pytest-embedded-qemu/pytest_embedded_qemu/qemu.py
```diff
@@ -41,13 +40,16 @@ def __init__(
...
+        super().__init__(
+            cmd=[qemu_prog_path, *qemu_cli_args, *qemu_extra_args] + ['-drive', f'file={image_path},if=mtd,format=raw'],
+            **kwargs,
+        )
 
-        cmd = f'{qemu_prog_path} {qemu_cli_args} {" ".join(qemu_extra_args)}'
-        logging.debug(cmd)
-
-        super().__init__(msg_queue, cmd, shell=True, **kwargs)
+    def _hard_reset(self):
```
hfudev, line 51: @igrr this is a drafted version. The un-implemented solution is to terminate the qemu process and re-create it with the same arguments.

Is there any better method to fake the `hard_reset` in qemu process?

igrr: We can also send `system_reset` command to QEMU monitor (console).  This will probably require routing the console to another destination, like a TCP port: `-monitor tcp::4444,server=on,wait=off`

(more options can be found here: https://qemu-project.gitlab.io/qemu/system/invocation.html)

igrr: And there's also QMP which is probably the better option, but I haven't personally used it yet: https://wiki.qemu.org/Documentation/QMP. There's even a Python package available to talk over QMP: https://pypi.org/project/qemu.qmp/.

hfudev: Thanks! I'll take a look.

hfudev: @igrr I added a tcp port as backend for QMP server. PTAL again!

hfudev:
@igrr PTAL the qmp `system_reset` solution. The reason I'm using unix socket instead of tcp connection is because there will be some `TIME_WAIT` socket connection left even I close the qmp client. That would somehow occupy the port and make the next test case failed while setup.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -351,12 +167,20 @@ def _parse_unity_menu_from_str(s: str) -> t.List[UnittestMenuCase]:
 
         return test_menu
 
+    def _hard_reset(self):
+        if hasattr(self, 'serial'):
+            return self.serial.hard_reset()
+        elif hasattr(self, 'qemu'):
+            return self.qemu._hard_reset()  # noqa
```
igrr, line 174: This looks a bit hacky, since this prevents new DUT classes being added without modifying this file. Plus, we can forget to modify this file when a new DUT is added...

Can we perhaps let the DUT register a callback in the mixin (`self._hard_reset_func`)?

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -527,6 +365,211 @@ def run_all_single_board_cases(
...
+    """
+    Dut class for serial ports connect to Espressif boards which are flashed with ESP-IDF apps
+
+    Attributes:
+        target (str): target chip type
+        skip_check_coredump (bool): skip check core dumped or not while dut teardown if set to True
+    """
+
+    XTENSA_TARGETS = ['esp32', 'esp32s2', 'esp32s3']
+    RISCV32_TARGETS = ['esp32c3', 'esp32h2', 'esp32c2']
```
igrr, line 379: c6 as well?

igrr:
Otherwise LGTM.

pytest-embedded-idf/pytest_embedded_idf/dut.py
hfudev: sure. applied with your suggestions

pytest-embedded-idf/pytest_embedded_idf/dut.py
hfudev: sure. also support detect with `IDF_TARGET_ARCH_XTENSA` and `IDF_TARGET_ARCH_RISCV`

hfudev:
@igrr I've removed the qmp support. it's really unstable.

Lots of exceptions will be thrown while deconstruction.

```
ERROR    qemu.qmp.protocol:protocol.py:910 Task.Writer: asyncio.exceptions.CancelledError
ERROR    qemu.qmp.protocol:protocol.py:910 Task.Reader: asyncio.exceptions.CancelledError
ERROR    qemu.qmp.protocol:protocol.py:910 Task.Writer: asyncio.exceptions.CancelledError
ERROR    qemu.qmp.protocol:protocol.py:910 Task.Reader: asyncio.exceptions.CancelledError
```

related issue: https://gitlab.com/qemu-project/python-qemu-qmp/-/issues/6

Shall we merge it for now? IMHO writing a `\n` for qemu dut hard reset is of course not a correct solution, but a okay solution though...

Issue #186, opened 2023-01-28
[idf] Support `sdkconfig.rename`


hfudev:
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig.html#backward-compatibility-of-kconfig-options

sometimes the setup period process depends on the kconfig item. Support the `sdkconfig.rename` for backward / forward compatibility.

PR #187, opened 2023-01-28
chore(idf): possible renaming in idf

hfudev:


pytest-embedded-idf/pytest_embedded_idf/app.py
```diff
@@ -52,7 +52,8 @@ def __init__(
         self.elf_file = self._get_elf_file()
 
         # loadable elf file skip the rest of these
-        if self.sdkconfig.get('APP_BUILD_TYPE_ELF_RAM'):
+        # TODO to be improved in #186
+        if self.sdkconfig.get('APP_BUILD_TYPE_ELF_RAM') or self.sdkconfig.get('APP_BUILD_TYPE_RAM'):
```
igrr, line 56: As an alternative, can we set app.is_loadable_elf via pytest.mark.parametrize in the test case itself?

hfudev: IMHO, I prefer #186 instead of setting it explicitly, since users may forget to set it again.

igrr: Actually, what is the point of #186? `sdkconfig` file already contains all the options added by sdkconfig.rename, so if `APP_BUILD_TYPE_ELF_RAM` got renamed into `APP_BUILD_TYPE_ELF`, then `sdkconfig` should contain both options.

igrr (approved):

pytest-embedded-idf/pytest_embedded_idf/app.py
hfudev: yes. sorry I didn't explain #186 clearly. The reason that we're having this issue is like you said, the sdkconfig.json doesn't contain the history names.

PR #202, opened 2023-04-20
pytest-embedded-qemu: creation of merged images if flash-encryption is enabled (RDT-443)

Harshal5:
pytest-embedded-qemu generates a merged image (bootloader + partition table + application) using the `make_bin` method of the class `IdfFlashImageMaker`.

In the case of flash encryption, the encrypted flag is set to True and the generation of the merged image fails. ([ref](https://github.com/espressif/pytest-embedded/blob/main/pytest-embedded-qemu/pytest_embedded_qemu/app.py#L34))

1. When flash encryption is enabled, in the case of the Self-Encryption workflow, the image is encrypted by the bootloader and so even if the `encrypted` flag is set, generating a merged image should be similar to that of the case when flash encryption is disabled.

2. When flash encryption is enabled, in the case of the Pre-Encryption workflow, the user must provide the encryption key file as an argument, and the above-generated merged image should be encrypted using the provided key.


pytest-embedded-qemu/pytest_embedded_qemu/qemu.py
```diff
@@ -15,7 +15,7 @@ class Qemu(DuplicateStdoutPopen):
     SOURCE = 'QEMU'
 
     QEMU_PROG_PATH = 'qemu-system-xtensa'
-    QEMU_DEFAULT_ARGS = '-nographic -no-reboot -machine esp32'
```
Harshal5, line 18: We need to remove the `-no-reboot` option from `QEMU_DEFAULT_ARGS`  as the bootloader calls a software reset after flash encryption has been completed.

hfudev:
@igrr PTAL!

hfudev:
This PR LGTM, but I know very little on the related areas, could you help take a look? @igrr

pytest-embedded-qemu/pytest_embedded_qemu/qemu.py
igrr: Is it possible to override flags specifically for the case where flash encryption workflow is being tested? I'm afraid this is a breaking change, otherwise.

igrr (approved):
Could you please add a test case to cover these changes?

igrr:
LGTM, just two requests (remove the breaking change and add a test, if possible.

pytest-embedded-qemu/pytest_embedded_qemu/qemu.py
hfudev: @igrr

I made the same change [here](https://github.com/espressif/pytest-embedded/pull/179/files#diff-a6ffa4515a4b71193b9b1eaff9e1fd25a01965cfd4632d75d99696c91cf6aa65R20) to enable the qmp `system_reset` function.

IMHO this breaking change is not "breaking" for testing purpose... since the side effect is call system_reset and show the output again. for me it's more like a fix. WDYT?

igrr: However if we had a test case which caused the emulated system to reset, thus finishing the test, will this test now just keep running? Do we currently assume that if QEMU process has exited, this means test failure?

igrr: Thinking about this again, i agree that this change is okay.

What i was worried about is that without -no-reboot we could fail to detect a crash in a hypothetical application which crashes on first boot but succeeds on the 2nd one.

But now i understand that we won't correctly detect such application as failing even when testing on real hardware, so the QEMU case is not anyhow "worse".

hfudev: > Do we currently assume that if QEMU process has exited, this means test failure?

no we don't. the pexpect process (also could call it the dut process) is reading the qemu output from the qemu process, and doing the expect function calls based on that.

pytest-embedded-qemu/tests/test_qemu.py
```diff
@@ -50,3 +50,33 @@ def test_pexpect_by_qemu(dut):
...
+
+    app_path = os.path.join(testdir.tmpdir, 'hello_world_esp32_flash_enc')
+    keyfile_path = os.path.join(app_path, 'pre_encryption_key.bin')
+    efuses_path = os.path.join(app_path, 'pre_encryption_efuses.bin')
+
+    result = testdir.runpytest(
+        '-s',
+        '--embedded-services', 'idf,qemu',
+        '--app-path', app_path,
+        '--qemu-extra-args',f'-drive file={efuses_path},if=none,format=raw,id=efuse\n'
```
hfudev, line 76: is the `\n` here on purpose?

tests/fixtures/hello_world_esp32_flash_enc/build/hello_world.elf
hfudev, line 1: is the elf file required? if not it's better to remove it, since it takes 2.34 MB.

hfudev (changes requested):
@Harshal5 Thanks! LGTM! left two minor comments.

hfudev (changes requested):
@Harshal5 Thanks! LGTM! left two minor comments.

tests/fixtures/hello_world_esp32_flash_enc/build/hello_world.elf
Harshal5: Right, removed it too.

Harshal5:
> @Harshal5 Thanks! LGTM! left two minor comments.

@hfudev PTAL, I have updated the MR.

hfudev:
LGTM! Thanks!

PR #204, opened 2023-05-10
Add boot loader entry in the flash args when secure boot is enabled

Harshal5:
`pytest-embedded-qemu` parses the `flasher_args.json` file [ref](https://github.com/espressif/pytest-embedded/blob/main/pytest-embedded-idf/pytest_embedded_idf/app.py#L188), to get the `FlashFiles` needed to generate a merged image using the `make_bin` method of the class `IdfFlashImageMaker`.

When secure boot is enabled, the bootloader entry is not included in the `flasher_args.json` file and thus the merged binary generated is bootloader-less.

This PR adds the following change in the `make_bin` method:
- Check if the config SECURE_BOOT is enabled in the `sdkconfig`.
- If enabled, append the bootloader entry and its corresponding offset in the flash into `flash_files`.


Harshal5:
@hfudev Could you PTAL? I am not sure why would the `Test Build Docs` check is failing.

hfudev:
> @hfudev Could you PTAL? I am not sure why would the `Test Build Docs` check is failing.

sure. I think the documentation is failing since griffe package `0.27.3`. Let's ignore this issue in your PR.

Harshal5:
> sure. I think the documentation is failing since griffe package `0.27.3`. Let's ignore this issue in your PR.

Oh okay, the PR is now ready to be reviewed. Could you PTAL?

hfudev:
@Harshal5 Hi, while I'm testing it locally, it takes me 300s to pass this test.

```
...
======================================================== 1 passed, 2 deselected, 1 warning in 299.26s (0:04:59) ========================================================
```

How about we double the timeout to 600s instead?

Besides of that, all LGTM!

Harshal5:
> How about we double the timeout to 600s instead?

Sure, I have updated the timeout to 600s.

igrr:
Wouldn't it be cleaner if we added a config option in IDF to "flash bootloader even if secure boot is enabled"? Otherwise we end up bringing knowledge of secure boot, bootloaders and other stuff into pytest-emebedded, which increases coupling between IDF and pytest-embedded.

Harshal5:
> Wouldn't it be cleaner if we added a config option in IDF to "flash bootloader even if secure boot is enabled", and enabled this option in test app config used with QEMU?
> 
> Otherwise we end up bringing knowledge of secure boot, bootloaders and other stuff into pytest-emebedded, which increases coupling between IDF and pytest-embedded.

Sure. The commit https://github.com/espressif/esp-idf/commit/873901e7aababb3e183766bd5f476eaddb8b1cc0 adds a config option in ESP-IDF to flash bootloader even if secure boot is enabled. Thus closing this PR.

Thank you!

PR #216, opened 2023-06-13
fix(qemu): qemu-system-xtensa < 8.0.0 only support 4MB flash size

hfudev:


hfudev:
related https://github.com/espressif/qemu/pull/68

PR #219, opened 2023-06-26
pytest_embedded_arduino/app.py: Add ESP32-C6 and ESP32-H2 to the list of supported chips 

P-R-O-C-H-Y:
Include ESP32-C6 to the list of supported chips in Arduino.


igrr:
@P-R-O-C-H-Y could we add esp32h2 as well, just so that we don't have to make a separate change for this?

P-R-O-C-H-Y:
> @P-R-O-C-H-Y could we add esp32h2 as well, just so that we don't have to make a separate change for this?

@igrr Done, but I am not sure about the offsets that's why I added only c6. Wanted to someone ask about that.

hfudev:
Thank you @P-R-O-C-H-Y!

PR #228, opened 2023-09-05
feat: Support hard_reset without stub (RDT-542)

Harshal5:
The method `hard_reset()` of the class `EspSerial` used the default arguments of the decorator use_esptool(), due to which stub is required for performing a hard_reset().

This support for `no_stub` allows usage of the package with the newer chips for whom `flasher_stub` support has not been added yet.

### DIFF
--- a/pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
+++ b/pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
@@ -202,7 +202,7 @@ class EspSerial(Serial):
     def _start(self):
         self.hard_reset()
 
-    @use_esptool()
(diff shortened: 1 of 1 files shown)


hfudev:
@dobairoland PTAL! I'm unaware of the side effects if we use no-stub mode for hard reset.

radimkarnis:
@dobairoland exactly. It is completely irrelevant for the hard reset if the stub is used or not.

`hard_reset` should just assert the DTR and RTS lines of the serial port, it doesn't communicate with the ESP in any way.

hfudev:
@dobairoland @radimkarnis Thank you for your explanation!

@Harshal5 Thank you for the PR! I'm merging it.

PR #229, opened 2023-09-21
feature: multi source expect (RDT-546)

horw:
Add functionality for expectations from many sources capability.
Created pexpect for many sources, add functionality for expect in DUT.


horw:
Hi, as I just add logic for DuplicateStdoutPopen child, to be exact create pexpect_process for them.
What do you think about this solution?
p.s. just implemented expect, still need update expect_exact etc.

PR #230, opened 2023-09-25
feature: Wokwi simulation support (RDT-551)

urish:
Functional implementation draft. Tested with https://github.com/espressif/gh-esp-test-template on simulated ESP32C6.

<strike>The Wokwi CLI still has no support for specifying multiple flash images. You can either give it an application binary file or elf file to load at 0x10000, using a default bootloader and partition table, or specify a full flash image to load. I'm still trying to figure out what's the best way to address it. Meanwhile, we only provide the app image to the CLI as a workaround.</strike>

Also, we can automatically download the CLI for the user, in case it's not installed. Not sure if this is a good idea, and where we should install it. It's distributed as a single binary, and currently [supports](https://github.com/wokwi/wokwi-cli/releases/tag/v0.6.7) Linux (arm, x64), MacOS (arm, x64) and Windows (x64). 

I'd appreciate your guidance on how to move forward with this.

Here's the output from the test setup I used, for reference:

```
(pytest) PS C:\p\gh-esp-test-template\test_app> pytest --target=esp32-c6 --embedded-services idf,wokwi
========================================================= test session starts ==========================================================
platform win32 -- Python 3.8.10, pytest-7.4.2, pluggy-1.3.0
rootdir: C:\p\gh-esp-test-template\test_app
configfile: pytest.ini
plugins: cov-4.1.0, embedded-1.3.5, rerunfailures-12.0
collected 1 item

test_app.py::test_app
------------------------------------------------------------ live log setup ------------------------------------------------------------ 
2023-09-25 23:11:24 INFO Executing wokwi-cli C:\p\gh-esp-test-template\test_app
2023-09-25 23:11:24 Wokwi CLI v0.6.7 (b824d8cb0edc)
2023-09-25 23:11:25 Connected to Wokwi Simulation API 1.0.0-20230922-g99ff9b94
2023-09-25 23:11:26 Starting simulation...
2023-09-25 23:11:27 ESP-ROM:esp32c6-20220919
2023-09-25 23:11:27
2023-09-25 23:11:27 Build:Sep 19 2022
2023-09-25 23:11:27 rst:0x1 (POWERON),boot:0x5c (SPI_FAST_FLASH_BOOT)
2023-09-25 23:11:27 SPIWP:0xee
2023-09-25 23:11:27 mode:DIO, clock div:2
2023-09-25 23:11:27 load:0x40875720,len:0x126c
2023-09-25 23:11:27 load:0x4086c410,len:0x704
2023-09-25 23:11:27 load:0x4086e610,len:0x2ae0
2023-09-25 23:11:27 SHA-256 comparison failed:
2023-09-25 23:11:27 Calculated: f5b31c96161053e152c0c7cb85092e2fd16385d8e5ebf950cfc21c815848383a
2023-09-25 23:11:27 Expected: 117a5a689f626f77701cc2df471cb76b8a20232ff4425b900e23bf4041d0c79b
2023-09-25 23:11:27 Attempting to boot anyway...
2023-09-25 23:11:27 entry 0x4086c410
2023-09-25 23:11:27 I (20) cpu_start: Unicore app
2023-09-25 23:11:27 I (20) cpu_start: Pro cpu up.
2023-09-25 23:11:27 W (10) clk: esp_perip_clk_init() has not been implemented yet
2023-09-25 23:11:27 I (11) cpu_start: Pro cpu start user code
2023-09-25 23:11:27 I (11) cpu_start: cpu freq: 160000000 Hz
2023-09-25 23:11:27
2023-09-25 23:11:27 I (11) cpu_start: Application information:
2023-09-25 23:11:27 I (11) cpu_start: Project name:     example_test_app
2023-09-25 23:11:27 I (11) cpu_start: App version:      681415e-dirty
2023-09-25 23:11:27 I (11) cpu_start: Compile time:     Sep 24 2023 12:07:49
2023-09-25 23:11:27 I (11) cpu_start: ELF file SHA256:  000000000...
2023-09-25 23:11:27 I (11) cpu_start: ESP-IDF:          v5.2-dev-2934-g3b748a6cb7
2023-09-25 23:11:28 I (12) cpu_start: Min chip rev:     v0.0
2023-09-25 23:11:28
2023-09-25 23:11:28 I (12) cpu_start: Max chip rev:     v0.99
2023-09-25 23:11:28 I (12) cpu_start: Chip rev:         v0.0
2023-09-25 23:11:28 I (12) heap_init: Initializing. RAM available for dynamic allocation:
2023-09-25 23:11:28 I (12) heap_init: At 4080BBD0 len 00070A40 (450 KiB): D/IRAM
2023-09-25 23:11:28 I (12) heap_init: At 4087C610 len 00002F54 (11 KiB): STACK/DIRAM
2023-09-25 23:11:28 I (13) heap_init: At 50000000 len 00003FE8 (15 KiB): RTCRAM
2023-09-25 23:11:28 I (13) spi_flash: detected chip: generic
2023-09-25 23:11:28 I (13) spi_flash: flash io: dio
2023-09-25 23:11:28 I (13) sleep: Configure to isolate all GPIO pins in sleep state
2023-09-25 23:11:28 I (14) sleep: Enable automatic switching of GPIO sleep configuration
2023-09-25 23:11:28 I (14) coexist: coex firmware version: 8770c12
2023-09-25 23:11:28 I (14) coexist: coexist rom version 5b8dcfa
2023-09-25 23:11:28 I (14) app_start: Starting scheduler on CPU0
2023-09-25 23:11:28
2023-09-25 23:11:28 I (14) main_task: Started on CPU0
2023-09-25 23:11:28 I (64) main_task: Calling app_main()
2023-09-25 23:11:28 Unity test run 1 of 1
2023-09-25 23:11:28 TEST(testable, mean_of_empty) PASS
2023-09-25 23:11:28 TEST(testable, mean_of_vector) PASS
2023-09-25 23:11:28 IGNORE_TEST(testable, test_fail)
2023-09-25 23:11:28
2023-09-25 23:11:28 -----------------------
2023-09-25 23:11:28 3 Tests 0 Failures 1 Ignored
2023-09-25 23:11:28 OK
2023-09-25 23:11:28
2023-09-25 23:11:28 Tests finished, rc=0
2023-09-25 23:11:28 I (74) main_task: Returned from app_main()
PASSED
---------------------------------------------------------- live log teardown ----------------------------------------------------------- 
2023-09-25 23:11:28 INFO Created unity output junit report: C:\Users\Uri\AppData\Local\Temp\pytest-embedded\2023-09-25_20-11-24-643490\test_app\dut.xml


========================================================== 1 passed in 3.77s =========================================================== 
```

### DIFF
--- a/pytest-embedded-wokwi/README.md
+++ b/pytest-embedded-wokwi/README.md
@@ -1,1 +1,37 @@
 ### pytest-embedded-wokwi
+
+pytest-embedded service for running tests on [Wokwi](https://wokwi.com/ci) instead of the real target.
+
+Wokwi supports most ESP32 targets, including: esp32, esp32s2, esp32s3, esp32c3, esp32c6, and esp32h2. In addition, it supports a [wide range of peripherals](https://docs.wokwi.com/getting-started/supported-hardware), including sensors, displays, motors, and debugging tools.
+
+Running the tests with Wokwi requires an internet connection. Your firmware is uploaded to the Wokwi server for the duration of the simulation, but it is not saved on the server. On-premises Wokwi installations are available for enterprise customers.
+
+## Wokwi CLI installation
+
+The Wokwi plugin uses the [Wokwi CLI](https://github.com/wokwi/wokwi-cli) to interact with the wokwi simulation server. You can download the precompiled CLI binaries from the [releases page](https://github.com/wokwi/wokwi-cli/releases). Alternatively, on Linux or Mac OS, you can install the CLI using the following command:
+
+```bash
+curl -L https://wokwi.com/ci/install.sh | sh
+```
+
+And on Windows:
+
+```powershell
+iwr https://wokwi.com/ci/install.ps1 -useb | iex
+```
+
+## API Tokens
+
+Before using this plugin, you need to create a free Wokwi account and [generate an API key](https://wokwi.com/dashboard/ci). You can then set the `WOKWI_CLI_TOKEN` environment variable to the API key.
+
+Linux / Mac OS / WSL:
+
+```bash
+export WOKWI_CLI_TOKEN="your-api-key"
+```
+
+Windows PowerShell:
+
+```powershell
+$env:WOKWI_CLI_TOKEN="your-api-key"
+```
--- a/pytest-embedded-wokwi/README.md
+++ b/pytest-embedded-wokwi/README.md
@@ -1,0 +1,1 @@
+### pytest-embedded-wokwi
--- a/pytest-embedded-wokwi/pyproject.toml
+++ b/pytest-embedded-wokwi/pyproject.toml
@@ -6,6 +6,7 @@ build-backend = "flit_core.buildapi"
 name = "pytest-embedded-wokwi"
 authors = [
     {name = "Fu Hanxi", email = "fuhanxi@espressif.com"},
+    {name = "Uri Shaked", email = "uri@wokwi.com"},
 ]
 readme = "README.md"
 license = {file = "LICENSE"}
--- a/pytest-embedded-wokwi/pyproject.toml
+++ b/pytest-embedded-wokwi/pyproject.toml
@@ -1,0 +1,105 @@
+[build-system]
+requires = ["flit_core >=3.2,<4"]
+build-backend = "flit_core.buildapi"
+
+[project]
+name = "pytest-embedded-wokwi"
+authors = [
+    {name = "Fu Hanxi", email = "fuhanxi@espressif.com"},
+]
+readme = "README.md"
+license = {file = "LICENSE"}
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Framework :: Pytest",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python",
+    "Topic :: Software Development :: Testing",
+]
+dynamic = ["version", "description"]
+requires-python = ">=3.7"
+
+dependencies = [
+    "pytest-embedded~=1.3.5",
+]
+
+[project.optional-dependencies]
+idf = [
+    "pytest-embedded-idf~=1.3.5",
+]
+
+[project.urls]
+homepage = "https://github.com/espressif/pytest-embedded"
+repository = "https://github.com/espressif/pytest-embedded"
+documentation = "https://docs.espressif.com/projects/pytest-embedded/en/latest/"
+changelog = "https://github.com/espressif/pytest-embedded/blob/main/CHANGELOG.md"
+
+[tool.isort]
+profile = 'black'
+
+[tool.black]
+line-length = 120
+target-version = ['py37']
+force-exclude = '/tests/'
+skip-string-normalization = true
+
+[tool.ruff]
+select = [
+    'F',  # Pyflakes
+    'E',  # pycodestyle
+    'W',  # pycodestyle
+#    'C90',  # mccabe
+#    'I',  # isort
+#    'N',  # pep8-naming
+#    'D',  # pydocstyle
+#    'UP',  #  pyupgrade
+#    'YTT',  # flake8-2020
+#    'ANN',  # flake8-annotations
+#    'S',  # flake8-bandit
+#    'BLE',  # flake8-blind-except
+#    'FBT',  # flake8-boolean-trap
+#    'B',  # flake8-bugbear
+#    'A',  # flake8-builtins
+#    'COM',  # flake8-commas
+#    'C4',  #  flake8-comprehensions
+#    'DTZ',  # flake8-datetimez
+#    'T10',  # flake8-debugger
+#    'DJ',  #  flake8-django
+#    'EM',  #  flake8-errmsg
+#    'EXE',  # flake8-executable
+#    'ISC',  # flake8-implicit-str-concat
+#    'ICN',  # flake8-import-conventions
+#    'G',  # flake8-logging-format
+#    'INP',  # flake8-no-pep420
+#    'PIE',  # flake8-pie
+#    'T20',  # flake8-print
+#    'PYI',  # flake8-pyi
+#    'PT',  #  flake8-pytest-style
+#    'Q',  # flake8-quotes
+#    'RSE',  # flake8-raise
+#    'RET',  # flake8-return
+#    'SLF',  # flake8-self
+#    'SIM',  # flake8-simplify
+#    'TID',  # flake8-tidy-imports
+#    'TCH',  # flake8-type-checking
+#    'ARG',  # flake8-unused-arguments
+#    'PTH',  # flake8-use-pathlib
+#    'ERA',  # eradicate
+#    'PD',  #  pandas-vet
+#    'PGH',  # pygrep-hooks
+#    'PL',  #  Pylint
+#    'TRY',  # tryceratops
+#    'NPY',  # NumPy-specific rules
+#    'RUF',  # Ruff-specific rules
+]
+line-length = 120
+target-version = "py37"
--- a/pytest-embedded-wokwi/pyproject.toml
+++ b/pytest-embedded-wokwi/pyproject.toml
@@ -30,6 +30,7 @@ requires-python = ">=3.7"
 
 dependencies = [
     "pytest-embedded~=1.3.5",
+    "toml~=0.10.2",
 ]
 
 [project.optional-dependencies]
--- a/pytest-embedded-wokwi/pytest_embedded_wokwi/dut.py
+++ b/pytest-embedded-wokwi/pytest_embedded_wokwi/dut.py
@@ -1,0 +1,25 @@
+from typing import AnyStr
+
+from pytest_embedded.dut import Dut
+
+from .wokwi_cli import WokwiCLI
+
+
+class WokwiDut(Dut):
+    """
+    Wokwi DUT class
+    """
+
+    def __init__(
+        self,
+        wokwi: WokwiCLI,
+        **kwargs,
+    ) -> None:
+        self.wokwi = wokwi
+
+        super().__init__(**kwargs)
+
+        self._hard_reset_func = self.wokwi._hard_reset
+
+    def write(self, s: AnyStr) -> None:
+        self.wokwi.write(s)
--- a/pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
+++ b/pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
@@ -8,6 +8,8 @@ import toml
 from pytest_embedded import __version__
 from pytest_embedded.log import DuplicateStdoutPopen
 
+from .idf import IDFFirmwareResolver
+
 if t.TYPE_CHECKING:
     from pytest_embedded_idf.app import IdfApp
 
@@ -33,6 +35,7 @@ class WokwiCLI(DuplicateStdoutPopen):
 
     def __init__(
         self,
+        firmware_resolver: IDFFirmwareResolver,
         wokwi_cli_path: t.Optional[str] = None,
         app: t.Optional['IdfApp'] = None,
         **kwargs,
@@ -42,6 +45,7 @@ class WokwiCLI(DuplicateStdoutPopen):
             wokwi_cli_path: Wokwi CLI arguments
         """
         self.app = app
+        self.firmware_resolver = firmware_resolver
 
         self.create_wokwi_toml()
         self.create_diagram_json()
@@ -59,7 +63,7 @@ class WokwiCLI(DuplicateStdoutPopen):
 
     def create_wokwi_toml(self):
         app = self.app
-        flasher_args = Path(app.binary_path, 'flasher_args.json')
+        flasher_args = self.firmware_resolver.resolve_firmware(app)
         wokwi_toml_path = os.path.join(app.app_path, 'wokwi.toml')
         firmware_path = Path(flasher_args).relative_to(app.app_path).as_posix()
         elf_path = Path(app.elf_file).relative_to(app.app_path).as_posix()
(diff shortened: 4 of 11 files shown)


igrr:
Thanks for the PR @urish!

> You can either give it an application binary file or elf file to load at 0x10000, using a default bootloader and partition table, or specify a full flash image to load. I'm still trying to figure out what's the best way to address it.

Perhaps you can try to do it the same way this is handled in pytest-embedded-qemu, by creating a full flash image?

urish:
> Perhaps you can try to do it the same way this is handled in pytest-embedded-qemu, by creating a full flash image?

I considered that, but then realized it is probably beneficial to have an easy way to simulate any esp-idf project in Wokwi for VSCode and in general using the CLI. So solving this at pytest-embedded level won't benefit VSCode nor the CLI.

Internally, I've been using `idf.py uf2` to generate a single file with everything (bootloader, partitions, app) that I could load into the simulator. So that could have been another solution - but the UF2 file is not built by default, so it's one more step for users they could miss.

I'm trying to figure out what would be the most user friendly way for IDF users. Any thoughts?

igrr:
If you can support reading the information from `build/flasher_args.json` file generated by IDF, that could be a solution that both IDF and pytest-embedded could rely on. For pytest-embedded-idf, flasher_args.json is already handled [here](https://github.com/espressif/pytest-embedded/blob/57cfd9840402f243c39d105e62d45436d4b8ff8f/pytest-embedded-idf/pytest_embedded_idf/app.py#L221).

urish:
Right, added [support for reading flasher_args.json](https://github.com/wokwi/wokwi-cli/commit/8b653f152e2ceb9b29f192a41cd916095b471656#diff-8cbc4377ef1022725cf10b7446856c8ba561f524230385b7838d2aa64548ee17R13-R45). I like how this simplifies the code for the pytest driver.

pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
```diff
@@ -0,0 +1,85 @@
...
+        **kwargs,
+    ):
+        """
+        Args:
+            wokwi_cli_path: Wokwi CLI arguments
+        """
+        self.app = app
+        flasher_args = Path(app.binary_path, 'flasher_args.json')
+
+        with open(os.path.join(app.app_path, 'wokwi.toml'), 'wt') as f:
```
igrr, line 45: Is it possible to keep the auto-generated files in build directory, instead? (Or at least to check that the file exists and not overwrite it if it does.)

Another question, can we add an option to specify the TOML file? I imagine that for a project supporting multiple boards (e.g. ESP-BSP) we would want to use `@pytest.mark.parametrize` to specify different TOML files depending on the board.

pytest-embedded-wokwi/pytest_embedded_wokwi/dut.py
```diff
@@ -0,0 +1,25 @@
...
+        **kwargs,
+    ) -> None:
+        self.wokwi = wokwi
+
+        super().__init__(**kwargs)
+
+        self._hard_reset_func = self.wokwi._hard_reset
+
+    def write(self, s: AnyStr) -> None:
+        self.wokwi.write(s)
```
igrr, line 25: Is there a way to do anything else interesting with the `dut`, other than serial I/O?

For example if there is an LED in the diagram, can we get its state using something like `dut.get_board_state('led', 'level')`? Or, get the image from the emulated LCD? Or modify the state of a switch or another control on the board from the test case?

Not saying this is a must have to merge the PR, but I think it is features like this that would be unique for Wokwi and not available with the alternatives (real HW or QEMU).

pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
urish: I'm trying to figure out what's the best way to handle this. The tradeoff is between ease of getting started (minimal configuration), and compatibility with the VS Code plugin. 

The VS Code plugin expects both wokwi.toml and diagram.json to be in the same directory (you could have multiple directories with multiple configurations). 

For basic CLI usage, you could probably get away with just specifying the type of the target type and have the wokwi.toml / diagram.json generated as we do above, without committing it to the repo. This makes switching to a different target super easy, as pytest already knows about the target and can generate diagram.json as we're doing now. @SergioGasquez also mentioned to me he'd love to see something similar in the CLI (a flag that generates diagram.json given an MCU or so).

But once you want to get more out of the simulator, you'll probably want to commit diagram.json and possibly also wokwi.toml (e.g. if you use some [custom chips](https://docs.wokwi.com/vscode/project-config#custom-chips). In more complex use cases, switching to a different MCU probably means you need to adjust the connections anyway, so you'd want to have multiple diagram configurations, as you suggested.

So I'm really not sure how to go about this.

pytest-embedded-wokwi/README.md
```diff
@@ -0,0 +1 @@
+### pytest-embedded-wokwi
```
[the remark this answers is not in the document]
urish: 👍

WDYT about auto downloading the CLI in case it is not installed?

pytest-embedded-wokwi/pytest_embedded_wokwi/dut.py
urish: I'll look into this once we've figured out the other parts.

pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
igrr: What about supporting extra arguments in wokwi CLI:

* `--config=path` — specifies the config file, if not specified then wokwi.toml in the current working directory is used
* `--diagram=path` — specifies the diagram file, if not specified then diagram.json in the current working directory is used

pytest-embedded-qemu can then specify have config options to specify these files (including pytest.mark.parametrize), and this will still work in VS Code projects.

pytest-embedded-wokwi/README.md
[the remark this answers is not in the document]
igrr: Up to you — if you want to write code to handle downloading, i think we can merge it. I would probably keep it simple in pytest-embedded and ask the developers to include the emulator in their Docker images, instead.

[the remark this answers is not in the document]
urish: Updated README - added some background info, simplified CLI installation instructions, and explanation about getting a token

pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
urish: As a first step, we no longer override wokwi.toml or diagram.json if they already exist. 

If wokwi.toml already exist, we check if the firmware / elf paths are set correctly, and if they are not, we update them while keeping the rest of the config intact.

For diagram.json, if it already exists, we read it and check if the selected board matches the current target, and if it does not, we print a warning.

pytest-embedded-wokwi/pyproject.toml
```diff
@@ -0,0 +1,106 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python",
+    "Topic :: Software Development :: Testing",
+]
+dynamic = ["version", "description"]
+requires-python = ">=3.7"
+
+dependencies = [
+    "pytest-embedded~=1.3.5",
+    "toml~=0.10.2",
```
hfudev, line 33:
```suggestion
    "toml; python_version < '3.11'",
```

Since python 3.11, the `tomllib` is natively supported.

```python
def load_toml(filepath: t.Union[str, Path]) -> dict:
    try:
        import tomllib  # python 3.11

        try:
            with open(str(filepath), 'rb') as fr:
                return tomllib.load(fr)
        except Exception as e:
            raise InvalidTomlError(filepath, str(e))
    except ImportError:
        import toml

        try:
            return toml.load(str(filepath))
        except Exception as e:
            raise InvalidTomlError(filepath, str(e))
```

hfudev (commented):
Thank you for your PR!

I'd like to comment a bit on the project layout.

### `pytest-embedded-wokwi` Dependencies

For now, wokwi service highly bundled with the idf. For example, in `create_wokwi_toml`, it depends on `flasher_args.json`. It makes the `WokwiCLI` hard to run without ESP-IDF binaries.

How about we make `pytest-embedded-wokwi` always depends on `pytest-embedded-idf`, not optionally? I prefer the original way you implement it, as `pytest-embedded-idf-wokwi`

### `WokwiDut` Inheritance

In the `IdfUnityDutMixin`, we provide some functions to interact with the c unity framework. If `WokwiDut` can also inherit from this mixin class, the users could also use run c test cases on wokwi. That would make it even better :)

Overall your PR LGTM! Thanks again!

pytest-embedded-wokwi/pyproject.toml
urish: Thanks for the suggestion, however, I see that `tomllib` only includes the `load()` and `loads()` methods, but there is no way to write toml files. Any suggestions how to go about this?

hfudev: ah, sorry I didn't notice the call to the `write()`. The `tomllib` is only used for parsing. Let's keep it as what it is right now.

urish: Thanks for reviewing my PR!

Regarding esp-idf - is there currently a way to use pytest embedded without esp-idf? I see there's also an Arduino plugin. We can possibly support the other use cases as well, I'm just not sure if it's worth the effort right now. WDYT?

Regarding `IdfUnityDutMixin` - I think this already works? We're adding the mix in [here](https://github.com/espressif/pytest-embedded/pull/230/files#diff-b5a6b0a6e20cb69a3ce927b538961a62b3a189870fb359992536e7cc2618cfa5R1228-R1231). Not sure if this is the best way to go about this, I just copied from the qemu case.

hfudev: Yes, there is. For example, you may use `qemu` service alone, or `serial` service alone, to interact with a Qemu process, or with a `serial` object.

Regarding the `IdfUnityDutMixin`, yes, you've added the mixin when `idf` service is enabled. The situation here is slightly different. Unlike `qemu`, `wokwi` shouldn't be run alone, so inherit the mixin directly would more intuitive. This was my original idea.

Thanks for reminding me, if we're going to support arduino in the same package as well, I agree to use the name `pytest-embedded-wokwi`. We only need to rename the `WokwiCLI` to `WokwiIdfCLI` or something, to make the hard-coded `create_wokwi_toml` to IDF-specific.

So, in short... considering the future support to arduino, here's my opinions
- keep the current package name
- limit the service `wokwi`, it should be used together with `idf` (or `arduino` in the future)

  ```python
            if 'wokwi' in _services:
                from pytest_embedded_wokwi import WokwiDut

                classes[fixture] = WokwiDut
                kwargs[fixture].update(
                    {
                        'wokwi': None,
                    }
                )

                if 'idf' in _services:
                    from pytest_embedded_idf.unity_tester import IdfUnityDutMixin

                    mixins[fixture].append(IdfUnityDutMixin)
                else:
                    # - you may also declare a new type of error as well
                    # - or set the `idf` as a default fallback, if not exists
                    raise SystemExit('wokwi service should be used together with idf service')
  ```

- put the current `create_wokwi_toml` in IDF-specific class, use this class to initialize the `wokwi_cli` fixture when `idf` service is enabled. (we may use something else when `arduino` service is enabled in the future)

urish: Thanks, I will look into doing these changes tomorrow. Other than `idf` and `arduino`, are there any other plugins planned?

What wokwi cli essentially needs is a list of bin files, their offsets, and and elf file if exists. If there is a common way to get the list of bin files and their offset, perhaps we can make it framework agnostic?

urish: Thanks again for the detailed feedback, I pushed an update limiting `wokwi` to be used together with `idf`, as well as moving the IDF specific logic into a separate class.

pytest-embedded-wokwi/pyproject.toml
```diff
@@ -0,0 +1,106 @@
+[build-system]
+requires = ["flit_core >=3.2,<4"]
+build-backend = "flit_core.buildapi"
+
+[project]
+name = "pytest-embedded-wokwi"
+authors = [
+    {name = "Fu Hanxi", email = "fuhanxi@espressif.com"},
```
hfudev, line 8: This is your work. You could put your name here.

hfudev (approved):
LGTM! Thank you for your PR.

could you help create one simple test case? a hello world test case would be enough.

hfudev:
@urish Thank you for adding the test. Now everything runs perfectly.

One last thing, could you help add `pytest-embedded-wokwi` after https://github.com/espressif/pytest-embedded/blob/main/foreach.sh#L12 as well? It's a helpful dev script to install all packages.

urish:
Done! Locally, I'm developing on Windows, so I converted that script to Python, to make it cross platform. If you are interested, here's the python version that I'm using:

```python
#!/usr/bin/env python3

import sys
import subprocess
import os

DEFAULT_PACKAGES = [
    "pytest-embedded",
    "pytest-embedded-serial",
    "pytest-embedded-serial-esp",
    "pytest-embedded-idf",
    "pytest-embedded-jtag",
    "pytest-embedded-qemu",
    "pytest-embedded-arduino",
    "pytest-embedded-wokwi"
]

def run_command(command):
    try:
        subprocess.check_call(command, shell=True)
    except subprocess.CalledProcessError as e:
        print(f"An error occurred: {e}")
        return 1
    return 0

def main(action):
    res = 0

    # one-time command
    res |= run_command("pip install -U pip")
    if action in ["install-editable", "build", "publish"]:
        res |= run_command("pip install -U flit")

    for pkg in DEFAULT_PACKAGES:
        os.chdir(pkg)
        if action == "install-editable":
            res |= run_command("flit install -s")
        elif action == "install":
            res |= run_command("pip install .")
        elif action == "uninstall":
            res |= run_command(f"pip uninstall -y {pkg}")
        elif action == "build":
            res |= run_command("flit build")
        elif action == "publish":
            res |= run_command("flit publish")
        else:
            print("invalid argument. valid choices: install-editable/install/uninstall/build/publish")
            return 1
        os.chdir("..")

    return res

if __name__ == "__main__":
    action = sys.argv[1] if len(sys.argv) > 1 else "install"
    exit_code = main(action)
    sys.exit(exit_code)
```

pytest-embedded-wokwi/pytest_embedded_wokwi/dut.py
hfudev: created #233 to track it

hfudev:
@urish Thanks again for your great PR!

PR #234, opened 2023-10-24
fix(serial.py): Pytest-embedded leaks file descriptors when used with pytest-lazy-fixture (RDT-577)

### LINKED ISSUE #232: Pytest-embedded leaks file descriptors when used with pytest-lazy-fixture (RDT-557)
**Describe the bug**
Running pytest-embedded-serial and dividing the serial lines using fixtures and pytest-lazy-fixture leaves files open so eventually either OS or python runs out of file descriptors.

I have included a test example that reproduces the problem. Essentially I have a multi-dut test system where I detect and divide the connected duts into single dut references using fixtures. I can then run same tests to multiple duts or dut combinations by parametrizing them, but that requires lazy-fixtures.

It looks like at least one issue is with pytest_embedded_serial/serial.py, for some reason the multiprocessing.MessageQueue is not empty when 'end' command is processed so it's left hanging. I have attached a patch file that for me improves the situation somewhat so I can run more tests before eventual failure. There seems to be some other place as well that leaks files.

**To Reproduce**
1. unzip test_alot.zip
2. install pytest-lazy-fixture
3. Run `pytest ...`

**Expected behavior**
Tests should be able to run, test framework should clear all open files properly.

**Dev Environment (please complete the following information):**
 - OS: Debian Linux
 - Python Version: 3.9.2
 - Pytest Version 7.4.2
 - pytest-embedded Version 1.3.5
 - pytest-lazy-fixture Version: 0.6.3

[improvement.patch](https://github.com/espressif/pytest-embedded/files/12857476/improvement.patch)
[test_alot.zip](https://github.com/espressif/pytest-embedded/files/12857477/test_alot.zip)

horw:
Running pytest with these arguments, along with pytest-lazy-fixture, can lead to a file descriptor error: 'ValueError: filedescriptor out of range in select()'.
-s --count 2 --embedded-services serial --port  /dev/ttyUSB0|/dev/ttyUSB1 --baud=115200|115200

The current commit helps alleviate this error, but if there are a large number of tests, the file descriptor error still exists.
close #232

### DIFF
--- a/pytest-embedded/pytest_embedded/plugin.py
+++ b/pytest-embedded/pytest_embedded/plugin.py
@@ -2,6 +2,7 @@ import contextlib
 import datetime
 import dbm
 import functools
+import gc
 import importlib
 import io
 import logging
@@ -431,6 +432,16 @@ def multi_dut_generator_fixture(
                 logging.debug('%s: %s', obj, str(e))
                 return  # swallow up all error
             finally:
+                referrers = gc.get_referrers(obj)
+                for _referrer in referrers:
+                    if isinstance(_referrer, list):
+                        for _i, val in enumerate(_referrer):
+                            if val is obj:
+                                _referrer[_i] = None
+                    elif isinstance(_referrer, dict):
+                        for key, value in _referrer.items():
+                            if value is obj:
+                                _referrer[key] = None
                 del obj
 
         if _COUNT == 1:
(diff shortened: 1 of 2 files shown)


pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -409,6 +410,17 @@ def _close_or_terminate(obj):
                 return
 
             try:
+                referrers = gc.get_referrers(obj)
```
hfudev, line 413: how about we move this code block before `del obj`

horw:
@hfudev test passed, but I commented out the lazy-fixture test because its dependencies were not installed.

hfudev:
LGTM. created #237 to for a follow-up cleanup task.

PR #236, opened 2023-11-07
ref: use esptool main instead other func

### LINKED ISSUE #235: Revise the Code Flashing Binaries to Target (RDT-576)
For now we're calling internal esptool functions while `flash`. Research if we could

1. get target

	```python
	esp: esptool.ESPLoader = esptool.get_default_connected_device(
	    ports,
	    port=port,
	    connect_attempts=3,
	    initial_baud=baud,
	    chip=esptool_target,
	)
	```

2. call `esptool.main(*args, esp=esp)` to use esptool cli functions

Besides, the `subprocess` call of esptool existing elsewhere could also be replaced by `esptool.main`

horw:
- This PR replaced the use of the esptool functions with `esptool.main`.
- Added the flag '--esp-flash-force' for 'force mode for esptool'.

closes #235

### DIFF
--- a/pytest-embedded-arduino/pytest_embedded_arduino/serial.py
+++ b/pytest-embedded-arduino/pytest_embedded_arduino/serial.py
@@ -51,6 +51,9 @@ class ArduinoSerial(EspSerial):
             flash_settings.append(f'--{k}')
             flash_settings.append(v)
 
+        if self.esp_flash_force:
+            flash_settings.append('--force')
+
         try:
             esptool.main(
                 ['--chip', self.app.target, 'write_flash', *flash_files, *flash_settings],
--- a/pytest-embedded-arduino/pytest_embedded_arduino/serial.py
+++ b/pytest-embedded-arduino/pytest_embedded_arduino/serial.py
@@ -2,7 +2,7 @@ import logging
 from typing import Optional
 
 import esptool
-from pytest_embedded_serial_esp.serial import EspSerial, EsptoolArgs
+from pytest_embedded_serial_esp.serial import EspSerial
 
 from .app import ArduinoApp
 
@@ -40,33 +40,21 @@ class ArduinoSerial(EspSerial):
         """
         Flash the binary files to the board.
         """
-        flash_files = [
-            (offset, open(path, 'rb')) for (offset, path, encrypted) in self.app.flash_files if not encrypted
-        ]
+        flash_files = []
+        for offset, path, encrypted in self.app.flash_files:
+            if encrypted:
+                continue
+            flash_files.extend((str(offset), path))
 
-        default_kwargs = {
-            'addr_filename': flash_files,
-            'encrypt_files': None,
-            'no_stub': False,
-            'compress': True,
-            'verify': False,
-            'ignore_flash_encryption_efuse_setting': False,
-            'erase_all': False,
-            'encrypt': False,
-            'force': False,
-            'chip': self.app.target,
-        }
-
-        default_kwargs.update(self.app.flash_settings[self.app.target])
-        flash_args = EsptoolArgs(**default_kwargs)
+        flash_settings = []
+        for k, v in self.app.flash_settings[self.app.target].items():
+            flash_settings.append(f'--{k}')
+            flash_settings.append(v)
 
         try:
-            self.stub.change_baud(self.esptool_baud)
-            esptool.detect_flash_size(self.stub, flash_args)
-            esptool.write_flash(self.stub, flash_args)
-            self.stub.change_baud(self.baud)
+            esptool.main(
+                ['--chip', self.app.target, 'write_flash', *flash_files, *flash_settings],
+                esp=self.esp,
+            )
         except Exception:
             raise
-        finally:
-            for _, f in flash_files:
-                f.close()
--- a/pytest-embedded-idf/pytest_embedded_idf/serial.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/serial.py
@@ -1,12 +1,11 @@
+import contextlib
 import hashlib
 import logging
-import os
 import tempfile
 from typing import Optional, TextIO, Union
 
 import esptool
-from pytest_embedded.log import live_print_call
-from pytest_embedded_serial_esp.serial import EspSerial, EsptoolArgs
+from pytest_embedded_serial_esp.serial import EspSerial
 
 from .app import IdfApp
 
@@ -84,30 +83,36 @@ class IdfSerial(EspSerial):
         if self.app.bin_file:
             bin_file = self.app.bin_file
         else:
-            live_print_call(
-                [
-                    'esptool.py',
-                    '--chip',
-                    self.app.target,
-                    'elf2image',
-                    self.app.elf_file,
-                    *self.app.write_flash_args,
-                ],
-                msg_queue=self._q,
-            )
+            with contextlib.redirect_stdout(self._q):
+                esptool.main(
+                    [
+                        '--chip',
+                        self.app.target,
+                        'elf2image',
+                        self.app.elf_file,
+                        *self.app.write_flash_args,
+                    ],
+                    esp=self.esp,
+                )
             bin_file = self.app.elf_file.replace('.elf', '.bin')
 
-        live_print_call(
-            [
-                'esptool.py',
-                '--chip',
-                self.app.target,
-                '--no-stub',
-                'load_ram',
-                bin_file,
-            ],
-            msg_queue=self._q,
-        )
+        with contextlib.redirect_stdout(self._q):
+            esptool.main(
+                [
+                    '--chip',
+                    self.app.target,
+                    '--no-stub',
+                    'load_ram',
+                    bin_file,
+                ],
+                esp=self.esp,
+            )
+
+    def _force_flag(self):
+        config = self.app.sdkconfig
+        if any((config.get('CONFIG_SECURE_FLASH_ENC_ENABLED', False), config.get('CONFIG_SECURE_BOOT', False))):
+            return ['--force']
+        return []
 
     @EspSerial.use_esptool()
     def flash(self) -> None:
@@ -122,59 +127,62 @@ class IdfSerial(EspSerial):
             logging.error('No flash settings detected. Skipping auto flash...')
             return
 
-        flash_files = [(file.offset, open(file.file_path, 'rb')) for file in self.app.flash_files if not file.encrypted]
-        encrypt_files = [(file.offset, open(file.file_path, 'rb')) for file in self.app.flash_files if file.encrypted]
-
-        nvs_file = None
-        try:
-            if self.erase_nvs:
-                address = self.app.partition_table['nvs']['offset']
-                size = self.app.partition_table['nvs']['size']
-                nvs_file = tempfile.NamedTemporaryFile(delete=False)
-                nvs_file.write(b'\xff' * size)
-                if not isinstance(address, int):
-                    address = int(address, 0)
-
-                if self.app.flash_settings['encrypt']:
-                    encrypt_files.append((address, open(nvs_file.name, 'rb')))
+        _args = []
+        for k, v in self.app.flash_args['extra_esptool_args'].items():
+            if isinstance(v, bool):
+                if k == 'stub':
+                    if v is False:
+                        _args.append('--no-stub')
+                elif v:
+                    _args.append(f'--{k}')
+            else:
+                _args.append(f'--{k}')
+                if k == 'after':
+                    _args.append('hard_reset')
                 else:
-                    flash_files.append((address, open(nvs_file.name, 'rb')))
+                    _args.append(str(v))
 
-            # write_flash expects the parameter encrypt_files to be None and not
-            # an empty list, so perform the check here
-            default_kwargs = {
-                'addr_filename': flash_files,
-                'encrypt_files': encrypt_files or None,
-                'no_stub': False,
-                'compress': True,
-                'verify': False,
-                'ignore_flash_encryption_efuse_setting': False,
-                'erase_all': False,
-                'force': False,
-            }
+        _args.append('write_flash')
 
-            default_kwargs.update(self.app.flash_settings)
-            default_kwargs.update(self.app.flash_args.get('extra_esptool_args', {}))
-            args = EsptoolArgs(**default_kwargs)
+        if self.erase_nvs:
+            esptool.main(
+                [
+                    'erase_region',
+                    str(self.app.partition_table['nvs']['offset']),
+                    str(self.app.partition_table['nvs']['size']),
+                ],
+                esp=self.esp,
+            )
+            self.esp.connect()
 
-            self.stub.change_baud(self.esptool_baud)
-            esptool.detect_flash_size(self.stub, args)
-            esptool.write_flash(self.stub, args)
-            self.stub.change_baud(self.baud)
+        encrypt_files = []
+        flash_files = []
+        for file in self.app.flash_files:
+            if file.encrypted:
+                encrypt_files.append(hex(file.offset))
+                encrypt_files.append(str(file.file_path))
+            else:
+                flash_files.append(hex(file.offset))
+                flash_files.append(file.file_path)
 
-            if self._meta:
-                self._meta.set_port_app_cache(self.port, self.app)
-        finally:
-            if nvs_file:
-                nvs_file.close()
-                try:
-                    os.remove(nvs_file.name)
-                except OSError:
-                    pass
-            for _, f in flash_files:
-                f.close()
-            for _, f in encrypt_files:
-                f.close()
+        if flash_files and encrypt_files:
+            _args.extend(flash_files)
+            _args.append('--encrypt-files')
+            _args.extend(encrypt_files)
+        else:
+            if flash_files:
+                _args.extend(flash_files)
+            else:
+                _args.append('--encrypt')
+                _args.extend(encrypt_files)
+
+        _args.extend(self.app.flash_args['write_flash_args'])
+        _args.extend(self._force_flag())
+
+        esptool.main(_args, esp=self.esp)
+
+        if self._meta:
+            self._meta.set_port_app_cache(self.port, self.app)
 
     @EspSerial.use_esptool()
     def dump_flash(
@@ -207,15 +215,12 @@ class IdfSerial(EspSerial):
         else:
             raise ValueError('You must specify "partition" or ("address" and "size") to dump flash')
 
-        content = self.stub.read_flash(_addr, _size)
         if output:
-            if isinstance(output, str):
-                os.makedirs(os.path.dirname(output), exist_ok=True)
-                with open(output, 'wb') as f:
-                    f.write(content)
-            else:
-                output.write(content)
+            esptool.main(['read_flash', str(_addr), str(_size), str(output)], esp=self.esp)
         else:
+            with tempfile.NamedTemporaryFile() as fp:
+                esptool.main(['read_flash', str(_addr), str(_size), fp.name], esp=self.esp)
+                content = fp.read()
             return content
 
     @EspSerial.use_esptool()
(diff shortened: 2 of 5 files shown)


horw:
Actually, Arduino means Arduino application. I initially thought it was a chip. 😊

horw:
Finally passed...

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -84,30 +85,28 @@ def load_ram(self) -> None:
...
                     self.app.target,
-                    'elf2image',
-                    self.app.elf_file,
-                    *self.app.write_flash_args,
+                    '--no-stub',
+                    'load_ram',
+                    bin_file,
                 ],
-                msg_queue=self._q,
             )
```
hfudev, line 109: esp=self.esp here?

otherwise the serial connection will be re-established.

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -122,59 +121,48 @@ def flash(self) -> None:
...
+
+        esptool.main(_args, esp=self.esp)
+
+        if nvs_file:
+            nvs_file.close()
+            try:
+                os.remove(nvs_file.name)
+            except OSError:
+                pass
+
```
hfudev, line 163: still need `finally` here. `esptool.main` may fail?

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -207,15 +195,12 @@ def dump_flash(
...
-            else:
-                output.write(content)
+            esptool.main(['read_flash', str(_addr), str(_size), str(output)], esp=self.esp)
         else:
+            output = 'read_flash.tmp'
+            esptool.main(['read_flash', str(_addr), str(_size), output], esp=self.esp)
+            content = open(output, 'rb').read()
             return content
 
     @EspSerial.use_esptool()
```
hfudev, line 206: is `use_esptool()` still useful here? shall we deprecate this function and stop using it in the codebase?

hfudev (approved):
LGTM! @radimkarnis PTAL! Thank you

pytest-embedded-idf/pytest_embedded_idf/serial.py
horw: Yes, I think we can try adding `self.esp` to all `esptool.main` functions.

pytest-embedded-idf/pytest_embedded_idf/serial.py
horw: Agree

pytest-embedded-idf/pytest_embedded_idf/serial.py
horw: I suggest it is still useful because it helps with redirecting the thread.

horw: But I'm also interested in the possibility of initializing `self.esp` only once in the `__init__` function: `self.esp = esptool.detect_chip(self.proc, self.baud)`.

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -133,7 +133,7 @@ def _post_init(self):
 
         super()._post_init()
 
-    def use_esptool(hard_reset_after: bool = True, no_stub: bool = False):
```
hfudev, line 163: breaking change...

we should deprecate this param if it's no longer useful.

hfudev: btw, is hard_reset_after still useful?

hfudev (changes requested):

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
horw: We can move it into `esptool.main`.

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -101,38 +99,13 @@ def __init__(
...
-                except (FatalError, OSError) as err:
-                    if port is not None:
-                        raise
-                    print(f'{each_port} failed to connect: {err}')
-                    if _esp:
-                        # ensure unused port is closed.
-                        _esp._port.close()
-                    _esp = None
-            esp = _esp
+            esp = esptool.get_default_connected_device(
```
hfudev, line 102: yeah, if we have `self.esp = ...` here, we don't have to `detect_chip` everytime in `use_esptool`.

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -45,6 +44,7 @@ class EspSerial(Serial):
     """
 
     ESPTOOL_DEFAULT_BAUDRATE = 921600
+    BASE_OPERATION = ['--after', 'hard_reset']
```
radimkarnis, line 47: Is this necessary? `--after hard_reset` is the default in esptool and doesn't need to be specified.

radimkarnis (approved):
The approach of using `esptool.main` with `esp=esptool.get_default_connected_device` seems like the correct approach to me. From the point of esptool this LGTM!

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
horw: Good to know 😄 I will remove it then

hfudev:
rebased to run the tests

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
horw: I also have a question about `esptool.get_default_connected_device`. 
I've noticed that if I don't call it every time after running `esptool.main`, I encounter a problem. 
Could you please explain why this happens and if there is any way to avoid calling get_default_connected_device every time?"

pytest-embedded-arduino/pytest_embedded_arduino/serial.py
```diff
@@ -40,33 +40,21 @@ def flash(self) -> None:
...
-            'ignore_flash_encryption_efuse_setting': False,
-            'erase_all': False,
-            'encrypt': False,
-            'force': False,
-            'chip': self.app.target,
-        }
-
-        default_kwargs.update(self.app.flash_settings[self.app.target])
-        flash_args = EsptoolArgs(**default_kwargs)
+        flash_setting = []
```
hfudev, line 49: nitpick

```suggestion
        flash_settings = []
```

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -207,15 +198,12 @@ def dump_flash(
...
-                os.makedirs(os.path.dirname(output), exist_ok=True)
-                with open(output, 'wb') as f:
-                    f.write(content)
-            else:
-                output.write(content)
+            esptool.main(['read_flash', str(_addr), str(_size), str(output)], esp=self.esp)
         else:
+            with tempfile.NamedTemporaryFile() as fp:
+                esptool.main(['read_flash', str(_addr), str(_size), fp.name], esp=self.esp)
+                content = fp.read()
```
hfudev, line 206: or use stringIO as a io buffer

hfudev (changes requested):
Got a few more questions. Thank you for the cleanup. Now it looks much better.

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -122,59 +123,49 @@ def flash(self) -> None:
             logging.error('No flash settings detected. Skipping auto flash...')
             return
 
-        flash_files = [(file.offset, open(file.file_path, 'rb')) for file in self.app.flash_files if not file.encrypted]
-        encrypt_files = [(file.offset, open(file.file_path, 'rb')) for file in self.app.flash_files if file.encrypted]
+        data = json.load(open(os.path.join(self.app.binary_path, './flasher_args.json')))
```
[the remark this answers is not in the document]
horw: Cool

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -172,27 +143,18 @@ def use_esptool(hard_reset_after: bool = True, no_stub: bool = False):
...
-                                self.stub = self.esp.run_stub()
-
-                            ret = func(self, *args, **kwargs)
-                        finally:
-                            if hard_reset_after:
-                                self.esp.hard_reset()
-
-                            self.proc.apply_settings(settings)
-
+                        self.esp.connect('hard_reset')
```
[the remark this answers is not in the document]
horw: It's true

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -202,16 +164,16 @@ def wrapper(self, *args, **kwargs):
     def _start(self):
         self.hard_reset()
 
-    @use_esptool(hard_reset_after=True, no_stub=True)
+    @use_esptool()
     def hard_reset(self):
         """Hard reset your espressif device"""
-        pass
+        self.esp.hard_reset()
```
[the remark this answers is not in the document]
horw: Oooh, I think that why are `pass` was here.

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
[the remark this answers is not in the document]
horw: I checked the code, and there are a lot of type resets. However, there is no 'hard_reset' here for connection function. You can try give it a random name, and it will call `class ClassicReset(ResetStrategy):` class anyway. But the `hard_reset` function will  call `class HardReset(ResetStrategy):`.

pytest-embedded-idf/pytest_embedded_idf/serial.py
horw: I used NamedTemporaryFile because esptool requires a file name.

hfudev: good to know. Thank you!

pytest-embedded-arduino/pytest_embedded_arduino/serial.py
```diff
@@ -40,33 +40,21 @@ def flash(self) -> None:
...
+            flash_settings.append(f'--{k}')
+            flash_settings.append(v)
 
         try:
-            self.stub.change_baud(self.esptool_baud)
-            esptool.detect_flash_size(self.stub, flash_args)
-            esptool.write_flash(self.stub, flash_args)
-            self.stub.change_baud(self.baud)
+            esptool.main(
+                ['--chip', self.app.target, 'write_flash', *flash_files, *flash_settings, '--force'],
```
hfudev, line 56: Hi @radimkarnis. Here the esptool commands are with `--force` option. We've tested in the internal CI, and the pipeline won't pass without them. could you help take a final review?

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -122,59 +121,62 @@ def flash(self) -> None:
...
+        else:
+            if flash_files:
+                _args.extend(flash_files)
+            else:
+                _args.append('--encrypt')
+                _args.extend(encrypt_files)
+
+        _args.extend(self.app.flash_args['write_flash_args'])
+
+        _args.append('--force')
```
hfudev, line 175: Here with `--force` as well

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -202,16 +167,20 @@ def wrapper(self, *args, **kwargs):
...
         logging.info('Erasing the flash')
-        self.stub.erase_flash()
+        esptool.main(['erase_flash', '--force'], esp=self.esp)
 
         if self._meta:
             self._meta.drop_port_app_cache(self.port)
+
+    @property
+    def stub(self):
+        warn('Use self.esp instead of self.stub. self.stub is deprecated.', DeprecationWarning)
```
hfudev, line 185:
```suggestion
        warn('Please use `self.esp` instead of `self.stub`. `self.stub` will be removed in 2.0 release.', DeprecationWarning)
```

hfudev (approved):
LGTM! Thank you @horw for this nice improvement.

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
horw: Updated

pytest-embedded-arduino/pytest_embedded_arduino/serial.py
radimkarnis: Hello guys,

`--force` is needed when:
1) The chip has flash encryption enabled
2) The chip has secure boot enabled
3) You are trying to flash an image for another chip revision (the chip revision is incompatible with the image)

ad 1) and 2) - I doubt you are using chips with active security features in the CI, so I guess the `--force` flag is needed for case 3). In that case, the better solution is to fix the image revision field in the image. Can you do that?

I am not a big fan of hard coding in the `--force` argument, it is there for a reason.

hfudev: @radimkarnis Thank you for the review. Now it's more clear to us.

For use case 1), we could improve it by

```python
if self.app.sdkconfig.get('CONFIG_SECURE_FLASH_ENC_ENABLED', False):
    args.append('--force')
```

For use case 2), we could improve it by

```python
if self.app.sdkconfig.get('CONFIG_SECURE_BOOT', False):
    args.append('--force')
```

as for use case 3), shouldn't we do it in `esptool` side? writing these logic inside `pytest-embedded` is quite weird. Or, is there any easier commands that we can call? if we could have `esptool flash --project-description-json <filepath>` that would be great.

I don't like the hard-coded `--force` either. For now I can suggest a quick workaround, let's add one `--esp-flash-force` CLI option insided `pytest-emebdded` for the replacement of use case 3). if this flag is specified, we'll add the `--force` while calling esptool flash.

horw: I have added the `--esp-flash-force` flag and a dependency on sdkconfig.

radimkarnis: ad 1) and 2) sound good to me! 

case 3): what exactly would you want to change in esptool? The correct behavior is to write correct tests (tests that use firmware images with the correct `min_rev` and `max_rev` fields - in this case, nothing needs to be changed in esptool and pytest-embedded).

But I am ok with the `--esp-flash-force` option, it adds flexibility.

LGTM!

radimkarnis (approved):

pytest-embedded-arduino/pytest_embedded_arduino/serial.py
hfudev: okay. I see. Thanks for the review Radim :)

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -122,59 +129,62 @@ def flash(self) -> None:
...
+                encrypt_files.append(hex(file.offset))
+                encrypt_files.append(str(file.file_path))
+            else:
+                flash_files.append(hex(file.offset))
+                flash_files.append(file.file_path)
+
+        if flash_files and encrypt_files:
+            _args.extend(flash_files)
+            _args.append('--encrypt-files')
+            _args.extend(encrypt_files)
```
hfudev, line 173:
```suggestion
            _args.extend([*flash_files, '--encrypt-files', *encrypt_files])
```

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -122,59 +129,62 @@ def flash(self) -> None:
...
+        else:
+            if flash_files:
+                _args.extend(flash_files)
+            else:
+                _args.append('--encrypt')
+                _args.extend(encrypt_files)
```
hfudev, line 179:
```suggestion
                _args.extend(['--encrypt', *encrypt_files])
```

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -122,59 +129,62 @@ def flash(self) -> None:
...
+
+        _args.extend(self.app.flash_args['write_flash_args'])
+        _args.extend(self._force_flag())
```
hfudev, line 182:
```suggestion
        _args.extend([*self.app.flash_args['write_flash_args'], *self._force_flag()])
```

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -172,27 +143,21 @@ def use_esptool(hard_reset_after: bool = True, no_stub: bool = False):
...
-                                self.stub = self.esp.run_stub()
-
-                            ret = func(self, *args, **kwargs)
-                        finally:
-                            if hard_reset_after:
-                                self.esp.hard_reset()
-
-                            self.proc.apply_settings(settings)
-
+                        self.esp.connect()
```
hfudev, line 158: do we need it?

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -84,30 +83,38 @@ def load_ram(self) -> None:
...
-                '--chip',
-                self.app.target,
-                '--no-stub',
-                'load_ram',
-                bin_file,
-            ],
-            msg_queue=self._q,
-        )
+    def _force_flag(self):
+        if self.app.esp_flash_force:
```
hfudev, line 112: since the `flash` function belongs to `EspSerial`, the `esp_flash_force` should goes together with `serial`.

Could you move it into the `EspSerial.__init__`?

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -122,59 +129,62 @@ def flash(self) -> None:
...
+                esp=self.esp,
+            )
+            self.esp.connect()
+
+        encrypt_files = []
+        flash_files = []
+        for file in self.app.flash_files:
+            if file.encrypted:
+                encrypt_files.append(hex(file.offset))
+                encrypt_files.append(str(file.file_path))
```
hfudev, line 165:
```suggestion
                encrypt_files.extend([hex(file.offset), str(file.file_path)])
```

nitpick

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -122,59 +129,62 @@ def flash(self) -> None:
...
+            else:
+                flash_files.append(hex(file.offset))
+                flash_files.append(file.file_path)
```
hfudev, line 168: same, nitpick

hfudev (changes requested):
Nice. Final round. One major comment and a few nitpicks.

besides, regarding the commit message:

let's keep it as this:

- refactor: call esptool.main() instead of implementing on our own (combining your first two commits)
- feat: support `--esp-flash-force` to run esptool.flash with the force flag

then the user could trace the changelog better

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
horw: After a reset without a connection, we are not able to do anything. We need to connect first; otherwise, esptool will raise an error like 'Received incorrect bytes'.

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -202,16 +169,23 @@ def wrapper(self, *args, **kwargs):
...
         """Hard reset your espressif device"""
-        pass
+        self.esp.hard_reset()
 
     @use_esptool()
     def erase_flash(self):
         """Erase the complete flash"""
         logging.info('Erasing the flash')
-        self.stub.erase_flash()
+        esptool.main(['erase_flash', '--force'], esp=self.esp)
```
horw, line 180: @hfudev Should we use `self.esp_flash_force` here too?

horw:
@hfudev Thanks for your advice on commit messages! The nitpick has already been fixed.

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
hfudev: yes, I agree.

hfudev:
LGTM! let's have a new release!

PR #240, opened 2023-11-14
feat: create dut-factory method (RDT-587)

### LINKED ISSUE #231: Make Create `Dut` Instance Easier (RDT-554)
For now if we're going to create a `Dut` instance manually, we have to declare a lot of things manually

- `MessageQueue`, to collect the serial output
- `pexpect_logfile`, to init the pexpect process
- `listener`, to redirect the serial output to the pexpect_logfile
- ...

Set default values for the `Dut` class, or provide a factory function, to help users create `Dut` instance easier.

horw:
- created factory class for dut
closes #231

### DIFF
--- a/pytest-embedded/pytest_embedded/dut_factory.py
+++ b/pytest-embedded/pytest_embedded/dut_factory.py
@@ -1,0 +1,744 @@
+import datetime
+import gc
+import io
+import logging
+import multiprocessing
+import os
+import subprocess
+import sys
+import typing as t
+from collections import defaultdict
+from pathlib import Path
+
+from pytest_embedded_idf import CaseTester
+
+if t.TYPE_CHECKING:
+    from pytest_embedded_idf import LinuxSerial
+    from pytest_embedded_idf.dut import IdfDut
+    from pytest_embedded_jtag import Gdb, OpenOcd
+    from pytest_embedded_qemu import Qemu
+    from pytest_embedded_serial import Serial
+    from pytest_embedded_wokwi import WokwiCLI
+
+from . import App, Dut
+from .log import MessageQueue, PexpectProcess
+from .utils import FIXTURES_SERVICES, ClassCliOptions, to_str
+
+
+def _drop_none_kwargs(kwargs: t.Dict[t.Any, t.Any]):
+    return {k: v for k, v in kwargs.items() if v is not None}
+
+
+_ctx = multiprocessing.get_context()
+_stdout = sys.__stdout__
+dut_global_index = 0
+
+PARAMETRIZED_FIXTURES_CACHE = {}
+"""
+This variable holds values that were used in 'parametrize_fixtures'.
+It helps to obtain the necessary information for a custom DUT, such as '_meta', '_services', and 'test_case_name'.
+"""
+
+
+def msg_queue_gn() -> MessageQueue:
+    return MessageQueue(ctx=_ctx)
+
+
+def _listen(q: MessageQueue, filepath: str, with_timestamp: bool = True, count: int = 1, total: int = 1) -> None:
+    _added_prefix = False
+    while True:
+        msg = q.get()
+        if not msg:
+            continue
+
+        with open(filepath, 'ab') as fw:
+            fw.write(msg)
+            fw.flush()
+
+        _s = to_str(msg)
+        if not _s:
+            continue
+
+        prefix = ''
+        if total > 1:
+            source = f'dut-{count}'
+        else:
+            source = None
+
+        if source:
+            prefix = f'[{source}] ' + prefix
+
+        if with_timestamp:
+            prefix = datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S') + ' ' + prefix
+
+        if not _added_prefix:
+            _s = prefix + _s
+            _added_prefix = True
+        _s = _s.replace('\r\n', '\n')  # remove extra \r. since multi-dut \r would mess up the log
+        _s = _s.replace('\n', '\n' + prefix)
+        if prefix and _s.endswith(prefix):
+            _s = _s.rsplit(prefix, maxsplit=1)[0]
+            _added_prefix = False
+
+        _stdout.write(_s)
+        _stdout.flush()
+
+
+def _listener_gn(msg_queue, _pexpect_logfile, with_timestamp, dut_index, dut_total) -> multiprocessing.Process:
+    os.makedirs(os.path.dirname(_pexpect_logfile), exist_ok=True)
+    kwargs = {
+        'with_timestamp': with_timestamp,
+        'count': dut_index,
+        'total': dut_total,
+    }
+
+    return _ctx.Process(
+        target=_listen,
+        args=(
+            msg_queue,
+            _pexpect_logfile,
+        ),
+        kwargs=_drop_none_kwargs(kwargs),
+    )
+
+
+def _pexpect_fr_gn(_pexpect_logfile, _listener) -> t.BinaryIO:
+    Path(_pexpect_logfile).touch()
+    _listener.start()
+    return open(_pexpect_logfile, 'rb')
+
+
+def pexpect_proc_fn(_pexpect_fr) -> PexpectProcess:
+    return PexpectProcess(_pexpect_fr)
+
+
+def _fixture_classes_and_options_fn(
+    _services,
+    # parametrize fixtures
+    app_path,
+    build_dir,
+    port,
+    port_location,
+    port_mac,
+    target,
+    beta_target,
+    baud,
+    skip_autoflash,
+    erase_all,
+    esptool_baud,
+    esp_flash_force,
+    part_tool,
+    confirm_target_elf_sha256,
+    erase_nvs,
+    skip_check_coredump,
+    panic_output_decode_script,
+    openocd_prog_path,
+    openocd_cli_args,
+    gdb_prog_path,
+    gdb_cli_args,
+    no_gdb,
+    qemu_image_path,
+    qemu_prog_path,
+    qemu_cli_args,
+    qemu_extra_args,
+    wokwi_cli_path,
+    wokwi_timeout,
+    skip_regenerate_image,
+    encrypt,
+    keyfile,
+    # pre-initialized fixtures
+    dut_index,
+    _pexpect_logfile,
+    test_case_name,
+    pexpect_proc,
+    msg_queue,
+    _meta,
+) -> ClassCliOptions:
+    classes: t.Dict[str, type] = {}
+    mixins: t.Dict[str, t.List[type]] = defaultdict(list)
+    kwargs: t.Dict[str, t.Dict[str, t.Any]] = defaultdict(dict)
+
+    for fixture in FIXTURES_SERVICES.keys():
+        if fixture == 'app':
+            kwargs['app'] = {'app_path': app_path, 'build_dir': build_dir}
+            if 'idf' in _services:
+                if 'qemu' in _services:
+                    from pytest_embedded_qemu import DEFAULT_IMAGE_FN, QemuApp
+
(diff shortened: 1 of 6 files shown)


horw:
Hi @hfudev, this is a very initial draft commit for the implementation of the Fabrica class. I have a few questions. I have only created the base Dut class for now, but I don't think it's enough for the users. Should we implement other Duts in the same manner? (for example, IdfDut, SerialDut, etc.) It is possible, but it might result in a lot of duplicated code.

horw:
As I see this task, I will split it into the following tasks:
- [ ] Move function logic under fixtures to `dut_factory`.
- [ ] Test the moved solution.
- [ ] Create a `DutFactory` class using these `dut_factory` functions.

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -0,0 +1,655 @@
...
+
+class DutFactory:
+    def __init__(self):
+        self.obj_stack = []
+
+    # The object has to be closed;
+    # otherwise, the process will still be here,
+    # but there are some problems with deletion, most likely because of the fork process.
+    # Call close instead delete.
+    def close(self):
```
horw, line 552: At first, I tried using __del__, but it seems there are some issues with the process. So changed to close

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -0,0 +1,655 @@
...
+            'keyfile': None,
+            # common
+            'test_case_name': buff_parametrized_fixtures['test_case_name'],
+            '_meta': buff_parametrized_fixtures['_meta'],
+            # pre-initialized fixtures
+            'dut_index': dut_index_global,
+            '_pexpect_logfile': _pexpect_logfile,
+            'pexpect_proc': pexpect_proc,
+            'msg_queue': msg_queue,
+        }
```
horw, line 630: Need to have some discuss about default values and how users can pass new values. I used the `buff_parametrized_fixtures` variable to store the previous fixture result.

horw:
It still seems rough, but it's starting to work. @hfudev, what do you think about doing it this way? Also, I commented on some issues here.

horw:
@hfudev, Hello, I have cleaned the history and it seems like the test was passed.

hfudev:
```python
# test_foo.py
from pytest_embedded.dut_factory import DutFactory


def test_foo():
    dut = DutFactory.create(app_path='components/console/test_apps/console', target='esp32c6')
    dut.run_all_single_board_cases()
```

unfortunately it's not working with error message

```
            _pexpect_logfile = os.path.join(
>               buff_parametrized_fixtures['_meta'].logdir, f'custom-dut-{dut_index_global}.txt'
            )
E           KeyError: '_meta'

```

Seems like the `set_buff_parametrized_fixtures` will be triggered in another fixture `parametrize_fixtures`. If I don't use any fixture, the `set_buff_parametrized_fixtures` will be skipped.

Besides, could you also add the `DutFactory` to `pytest_embedded/__init__.py` `__all__`?

horw:
@hfudev 
Thanks for your review! Your testcase was very useful!

This commit contains the following changes:
- `parametrize_fixtures` changed to be autouse, which helps initialize `set_buff_parametrized_fixtures`.
- Added `unit_tester`.
- Added a test for `esp,idf`.
- Call `DutFactory.close()` after every test, not the session.

Do you have any other testcase I can try to implement?

Further steps:
- [ ] Add the `DutFactory` to `pytest_embedded/__init__.py ` `__all__`.
- [ ] Check `dut_factory` again, as there was some conflict related to the update in `plugin.py` when rebasing (I have just accepted my changes).
- [ ] Check for possible issues with `dut-{number}`.

pytest-embedded-idf/pytest_embedded_idf/serial.py
hfudev, line 1: better to make this change in a separate PR to get release faster...

pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -161,6 +161,8 @@ def flash(self, app: Optional[IdfApp] = None) -> None:
                 else:
                     _args.append(str(v))
 
+        _args.extend(['--baud', '1000000'])
```
hfudev, line 164: let's keep the behavior the same as `idf.py flash`

```
❯ idf.py flash --help
Usage: idf.py flash [OPTIONS]

  Flash the project.

Options:
  -C, --project-dir PATH         Project directory.
  -D, --define-cache-entry TEXT  Create a cmake cache entry. This option can be used at most once either globally, or for one subcommand.
  -b, --baud INTEGER             Baud rate for flashing. It can imply monitor baud rate as well if it hasn't been defined locally. The
                                 default value can be set with the ESPBAUD environment variable. This option can be used at most once
                                 either globally, or for one subcommand.
```

i.e., if ESPBAUD set, use the set value. if not, use the suggested value 1000000

pytest-embedded-idf/pytest_embedded_idf/serial.py
horw: Yes, I will do :+1:

pytest-embedded-idf/pytest_embedded_idf/serial.py
horw: Agree

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -0,0 +1,682 @@
...
+        qemu_prog_path=None,
+        qemu_cli_args=None,
+        qemu_extra_args=None,
+        wokwi_cli_path=None,
+        wokwi_timeout=0,
+        skip_regenerate_image=None,
+        encrypt=None,
+        keyfile=None,
+    ):
+        layout = []
```
hfudev, line 589: for public APIs, let's write some documentations to tell users how to call it.

like 

```python
def create():
    """
    Create DUT object according to the arguments.

    Examples:
    >>> foo = DutFactory.create(embedded_services='idf,esp', app_path='path_to_hello_world')
    >>> foo.expect_exact('Hello world!')
    """
```

horw:
> for example, if we support `--esptool-version` for `EspSerial, how many files shall we touch? (we won't support that, just an example) could you help document it somewhere in the comments as well?

Hello, let's take `wokwi_timeout` as an example. This is the output from `git grep`. As a result, you need to make changes in `dut_factory` and the `plugin`.
```
pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py:41:        wokwi_timeout: t.Optional[int] = None,
pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py:57:        if (wokwi_timeout is not None) and (wokwi_timeout > 0):
pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py:58:            cmd.extend(['--timeout', str(wokwi_timeout)])
pytest-embedded/pytest_embedded/dut_factory.py:140:    wokwi_timeout,
pytest-embedded/pytest_embedded/dut_factory.py:291:                    'wokwi_timeout': wokwi_timeout,
pytest-embedded/pytest_embedded/dut_factory.py:584:        wokwi_timeout=0,
pytest-embedded/pytest_embedded/dut_factory.py:638:                'wokwi_timeout': wokwi_timeout,
pytest-embedded/pytest_embedded/plugin.py:950:def wokwi_timeout(request: FixtureRequest) -> t.Optional[str]:
pytest-embedded/pytest_embedded/plugin.py:952:    return _request_param_or_config_option_or_default(request, 'wokwi_timeout', None)
pytest-embedded/pytest_embedded/plugin.py:1010:    wokwi_timeout,
```

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -0,0 +1,682 @@
...
+            elif k == 'gdb':
+                kwargs[k] = gdb
+            elif k == 'qemu':
+                kwargs[k] = qemu
+            elif k == 'wokwi':
+                kwargs[k] = wokwi
+    return cls(**_drop_none_kwargs(kwargs), mixins=mixins)
+
+
+buff_parametrized_fixtures = {}
```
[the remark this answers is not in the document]
horw: For the first run, the buffer helps store some fixtures such as metadata, test case names, and services, etc., for reuse. It might be better to change the name to cache.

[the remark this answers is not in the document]
horw: Renamed to cache

pytest-embedded/pytest_embedded/dut_factory.py
horw: Updated

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -0,0 +1,725 @@
...
+        wokwi_cli_path=None,
+        wokwi_timeout=0,
+        skip_regenerate_image=None,
+        encrypt=None,
+        keyfile=None,
+    ):
+        """
+        Create a Device Under Test (DUT) object with customizable parameters.
+
+        Parameters:
```
hfudev, line 591: Here you're writing documentation in a non-standard way. This repo is using google docstring style. https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#:~:text=def%20function_with_pep484_type_annotations(,False%20otherwise.%0A%0A%20%20%20%20%22%22%22

besides, it's much better to write the types in the function signatures than in the docstrings.

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -0,0 +1,725 @@
...
+            elif k == 'gdb':
+                kwargs[k] = gdb
+            elif k == 'qemu':
+                kwargs[k] = qemu
+            elif k == 'wokwi':
+                kwargs[k] = wokwi
+    return cls(**_drop_none_kwargs(kwargs), mixins=mixins)
+
+
+parametrized_fixtures_cache = {}
```
hfudev, line 481: could you place all the global variables together at the top of this file? and make them uppercase, with some documentation explaining what are these global vars storing.

(increasing the readability)

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -0,0 +1,725 @@
...
+        layout = []
+        try:
+            global parametrized_fixtures_cache
+            msg_queue = msg_queue_gn()
+            layout.append(msg_queue)
+
+            _pexpect_logfile = os.path.join(
+                parametrized_fixtures_cache['_meta'].logdir, f'custom-dut-{dut_global_index}.txt'
+            )
+            logging.warning('You can get your custom DUT log file at the following path: %s.', _pexpect_logfile)
```
hfudev, line 641:
```suggestion
            logging.debug('You can get your custom DUT log file at the following path: %s.', _pexpect_logfile)
```

either debug or info is fine, but not a warning.

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -0,0 +1,725 @@
...
+                        _referrer[_i] = None
+            elif isinstance(_referrer, dict):
+                for key, value in _referrer.items():
+                    if value is obj:
+                        _referrer[key] = None
+        del obj
+
+
+class DutFactory:
+    obj_stack: t.ClassVar[t.List[t.List[t.Any]]] = []
```
hfudev, line 531:
```suggestion
    # stores the objects that required by each dut
    # [
    #   [app, serial, ..., dut]  # dut-0
    #   [app, serial, ..., dut]  # dut-1
    #   ...
    # ]
    obj_stack: t.ClassVar[t.List[t.List[t.Any]]] = []
```

hfudev (changes requested):
@horw Thanks for this great PR. Only some minor comments regarding to the code conventions. Let's merge it today!

PR #242, opened 2023-11-29
feat: added interrupt if one of the test was failed and ability to data transfer with signal (RDT-593)

horw:
1. If one test case fails, call for an interruption to stop other tests, with the message: "Some of the DUTs failed, so this DUT was interrupted."
2. The ability to send data to other DUTs with a signal, like [Signal Name][Signal data].

### DIFF
--- a/pytest-embedded-idf/tests/test_idf.py
+++ b/pytest-embedded-idf/tests/test_idf.py
@@ -546,23 +546,14 @@ def test_unity_test_case_runner(testdir):
     assert junit_report.attrib['skipped'] == '0'
     assert junit_report.attrib['tests'] == '3'
 
-    case_names_one_dev = [
-        'normal_case1',
-        'normal_case2',
-        'multiple_stages_test',
-    ]
-    case_names_multi_dev = [
-        'multiple_devices_test',
-    ]
-
-    one_dev_dut = ['dut-0']
-
-    required_names_one_dev = [f'{case_name}' for _ in one_dev_dut for case_name in case_names_one_dev]
-    required_names_multi_dev = [f'{case_name}' for case_name in case_names_multi_dev]
-
-    junit_case_names = [item.attrib['name'] for item in junit_report]
-
-    assert sorted(required_names_one_dev + required_names_multi_dev) == sorted(junit_case_names)
+    assert junit_report[0].get('name') == 'normal_case1'
+    assert junit_report[0].find('failure') is None
+    assert junit_report[1].get('name') == 'normal_case2'
+    assert junit_report[1].find('failure') is not None
+    assert junit_report[2].get('name') == 'multiple_stages_test'
+    assert junit_report[2].find('failure') is None
+    assert junit_report[3].get('name') == 'multiple_devices_test'
+    assert junit_report[3].find('failure') is None
 
 
 def test_erase_all_with_port_cache(testdir):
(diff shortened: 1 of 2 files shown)


horw:
I investigated the testcase name, such as "test ROC and Offchannel Action Frame Tx [dut-0]". Is it important to include [dut-0] in the name?

hfudev:
> @hfudev, I investigated the testcase name, such as "test ROC and Offchannel Action Frame Tx [dut-0]," and then I have a small question. Is it important to include [dut-0] in the name? If the file name contains "dut-X," can I simply skip it, or will it break something?
> 
> [`a101546`/pytest-embedded/pytest_embedded/unity.py#L278C17-L278C17](https://github.com/espressif/pytest-embedded/blob/a10154665a28d3819f4460c2e346ff14a2d7e71a/pytest-embedded/pytest_embedded/unity.py#L278C17-L278C17)

The test case name should be `test ROC and Offchannel Action Frame Tx`

pytest-embedded-idf/tests/test_idf.py
```diff
@@ -556,10 +556,9 @@ def test_unity_test_case_runner(unity_tester):
...
 
     one_dev_dut = ['dut-0']
-    multi_dev_duts = ['dut-0', 'dut-1']
 
-    required_names_one_dev = [f'{case_name} [{dut}]' for dut in one_dev_dut for case_name in case_names_one_dev]
-    required_names_multi_dev = [f'{case_name} [{multi_dev_duts[0]}]' for case_name in case_names_multi_dev]
+    required_names_one_dev = [f'{case_name}' for _ in one_dev_dut for case_name in case_names_one_dev]
+    required_names_multi_dev = [f'{case_name}' for case_name in case_names_multi_dev]
 
     junit_case_names = [item.attrib['name'] for item in junit_report]
```
hfudev, line 563: could you help rewrite the test case a bit. assert the passed ones and the failed ones with their names.

assert junit_report.nodes[0] == 'normal_case1'
assert junit_report.nodes[1] == 'normal_case2'
assert 'failures' in junit_report.nodes[1]
assert junit_report.nodes[2] == 'multiple_stages_test'
assert junit_report.nodes[3] == 'multiple_devices_test'

pytest-embedded/tests/test_base.py
```diff
@@ -426,8 +426,8 @@ def test_expect_unity_test_output_multi_dut_record_1(dut):
         'test case 4',
     ]
     required_names = []
-    for dut in ['dut-0', 'dut-1', 'dut-1']:
-        required_names.extend([f'{case_name} [{dut}]' for case_name in case_names])
+    for _ in ['dut-0', 'dut-1', 'dut-1']:
+        required_names.extend([f'{case_name}' for case_name in case_names])
```
hfudev, line 430: likewise

hfudev (approved):
The library code LGTM! We should also optimize the test code a bit :)

pytest-embedded-idf/tests/test_idf.py
horw: It seems like 'nodes' is not the right solution. 
```AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'nodes'```
I will check how to implement it in other way.

File: pytest-embedded/tests/test_base.py [hunk 3]
```diff
@@ -473,8 +486,8 @@ def test_expect_unity_test_output_multi_dut_with_illegal_chars(dut):
     assert junit_report.attrib['tests'] == '2'
 
     required_names = [
-        'test_case [dut-0]',
-        'test_case [dut-1]',
+        'test_case',
+        'test_case',
```
hfudev, line 490: same here

PR #243, opened 2023-12-01
feat: tests with extra plugins (RDT-603)

### LINKED ISSUE #237: Rewrite Tests with Extra Plugins (RDT-580)
By default, we could enable plugins like `pytest-cov`, `pytest-rerunfailures` by default

Create a new folder and run with extra plugins for test cases that requires interactions with those plugins, like `pytest-lazy-fixture`

horw:
Created new folder for "tests with extra plugins"
closes #237


horw:
Just tested again, for me, it doesn't work if I start with testdir.makepyfile test. If I start directly, it will work properly.

PR #248, opened 2023-12-22
fix(unity): make single board test procedure more robust

hfudev:


pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -219,16 +221,28 @@ def _parse_unity_menu_from_str(s: str) -> t.List[UnittestMenuCase]:
 
         return test_menu
 
+    def _hard_reset(self) -> None:
+        if self._hard_reset_func:
+            try:
+                self._hard_reset_func()
+            except NotImplementedError:
+                # since menu printed but been expected (buffer has gone)... ignore the first ready pattern
```
horw, line 229: This is very useful!
By the way, I think we don't need the `self._ignore_first_ready_pattern` attribute, do we?
We can simply press enter every time before expecting. As I see it, it will work the same way.
```python
def _hard_reset(self) -> None:
    if self._hard_reset_func:
        try:
            self._hard_reset_func()
        except NotImplementedError:
            pass
def _get_ready(self, timeout: float = 30) -> None:
   self.confirm_write('\n', expect_pattern=READY_PATTERN_LIST, timeout=timeout)
```

hfudev: yes you're right. the downside of this approach is that, the test menu will be printed again after a reset.

hfudev:
Since the error that this PR is fixing is not 100% reproducible, I'll merge it first :D

PR #250, opened 2023-12-26
fix(serial): only delay on redirect thread with low baud rate (RDT-629)

thanhlev:


hfudev:
Hi @thanhlev, thank you for your PR! Could you provide more detail on the motivation behind this change?

The original reason for setting the read interval to 0.05 seconds was to prevent the read thread defined here from being executed too frequently. If this thread is executed too often, it may not have enough time to execute functions on your custom threads, if you're using multi-threading in you test scripts.

thanhlev:
> Hi @thanhlev, thank you for your PR! Could you provide more detail on the motivation behind this change?
> 
> The original reason for setting the read interval to 0.05 seconds was to prevent the read thread defined here from being executed too frequently. If this thread is executed too often, it may not have enough time to execute functions on your custom threads, if you're using multi-threading in you test scripts.

Hi @hfudev , Thanks for your reply.
I ran into an issue with baud rate 3M (missed data and NULL character on received data unexpectedly ), with baud 200K - got NULL character on received data unexpectedly.

The change will only target high baud rate testing

hfudev:
@thanhlev Could you help post the error message here and provide a minimal reproducible code snippet? Let's see if there's another way to fix it.

The current fix by your PR would cause problems when used together with multiple threads.

PR #251, opened 2023-12-29
fix(idf): make LinuxDut inherit from IdfUnityDutMixin when `idf` service enabled

hfudev:


hfudev:
@ESP-Marius PTAL as well

PR #258, opened 2024-01-09
feat: expect not matching keyword support

### LINKED ISSUE #257: Support `expect` With `not_matching` (RDT-636)
For example

```python
dut.expect('foo', not_matching='bar')
```

Will fail

```
bar foo
```

But pass

```
foo bar
```

check the `pexpect_proc.buffer` after a success match would be a quick solution.

horw:
Support for a `not_matching` argument has been added to the `expect` function. The function now utilizes `re.search()` to search for matches of the encoded keywords in `self.pexpect_proc.before`.

Closes #257


pytest-embedded/pytest_embedded/dut.py
```diff
@@ -117,7 +118,19 @@ def expect(self, pattern, **kwargs) -> Match:  # noqa
             - `AnyStr`: if you're matching `pexpect.EOF` or `pexpect.TIMEOUT` to get all the current buffers.
             - `re.Match`: if matched given string.
         """
-        return self.pexpect_proc.expect(pattern, **kwargs)
+
+        not_matching: List[str] = kwargs.pop('not_matching', [])
```
hfudev, line 122: Please also note it in the docstring.

pytest-embedded/pytest_embedded/dut.py
```diff
@@ -117,7 +118,19 @@ def expect(self, pattern, **kwargs) -> Match:  # noqa
...
+        index: Match = self.pexpect_proc.expect(pattern, **kwargs)
```
hfudev, line 123:
```suggestion
        index: int = self.pexpect_proc.expect(pattern, **kwargs)
```

index should be an int?

hfudev (changes requested):
one general question, shall we implement it in `_pexpect_func` decorator?

`expect_exact` should also have this kwargs support.

horw:
> one general question, shall we implement it in `_pexpect_func` decorator?
> 
> `expect_exact` should also have this kwargs support.

I think it is a good idea!

pytest-embedded/pytest_embedded/dut.py
horw: Yes, it returns an int, but I am just following the function annotation...
`def expect(self, pattern, **kwargs) -> Match:  # noqa`
Should we just change it to int, or is there some idea behind this?

pytest-embedded/pytest_embedded/dut.py
horw: Updated

pytest-embedded/pytest_embedded/dut.py
```diff
@@ -70,6 +71,7 @@ def wrapper(
         ) -> Union[Union[Match, AnyStr], List[Union[Match, AnyStr]]]:
             patterns = to_list(pattern)
             res = []
+            not_matching: List[str] = kwargs.pop('not_matching', [])
```
hfudev, line 74: let's move it to the function signature.

pytest-embedded/pytest_embedded/dut.py
```diff
@@ -80,11 +82,18 @@ def wrapper(
...
                 else:
-                    if self.pexpect_proc.match in [pexpect.EOF, pexpect.TIMEOUT]:
-                        res.append(self.pexpect_proc.before.rstrip())
-                    else:
-                        res.append(self.pexpect_proc.match)
+                    res.append(self.pexpect_proc.match)
+
+                if not isinstance(not_matching, list):
+                    not_matching = [not_matching]
+                for pattern in not_matching:
```
hfudev, line 93:
```suggestion
                for pattern in to_list(not_matching):
```

pytest-embedded/pytest_embedded/dut.py
```diff
@@ -131,6 +141,7 @@ def expect_exact(self, pattern, **kwargs) -> Match:  # noqa
             timeout (float): would raise `pexpect.TIMEOUT` exception when pattern is not matched after timeout
             expect_all (bool): need to match all specified patterns if this flag is `True`.
                 Otherwise match any of them could pass
+            not_matching (List[str]): patterns that should not appear in the buffer before expected patterns
```
hfudev, line 144:
```suggestion
            not_matching: string, or compiled regex, or a list of string and compiled regex.
```

The `not_matching` keyword should accept all these three types of arguments.

hfudev (changes requested):
LGTM! sorry for the late review.

pytest-embedded/pytest_embedded/dut.py
```diff
@@ -80,11 +81,20 @@ def wrapper(
...
+                    res.append(self.pexpect_proc.before.rstrip())
                 else:
-                    if self.pexpect_proc.match in [pexpect.EOF, pexpect.TIMEOUT]:
-                        res.append(self.pexpect_proc.before.rstrip())
-                    else:
-                        res.append(self.pexpect_proc.match)
+                    res.append(self.pexpect_proc.match)
+
+                for nm_pattern in to_list(not_matching):
+                    if isinstance(nm_pattern, str):
```
horw, line 91: Now it is possible to accept a `string`, `compiled regex`, or a `list` of strings and compiled regexes for our pattern.

horw:
I have updated this PR. Let's see the tests turn green!

pytest-embedded/pytest_embedded/dut.py
```diff
@@ -80,11 +81,20 @@ def wrapper(
...
+                        nm_pattern = re.compile(nm_pattern.encode())
+                    if isinstance(nm_pattern.pattern, str):
+                        nm_pattern = re.compile(nm_pattern.pattern.encode())
```
hfudev, line 94:
```suggestion
                    if isinstance(nm_pattern, re.Pattern):
                    	pass
```

otherwise it will be wrong when user is passing a `re.compile(b'xxx')`

pytest-embedded/pytest_embedded/dut.py
```diff
@@ -80,11 +81,20 @@ def wrapper(
...
+
+                    if nm_pattern.search(self.pexpect_proc.before):
+                        raise ValueError(f'The pattern {nm_pattern} should not have been matched.')
```
hfudev, line 97:
```suggestion
                        raise ValueError(f'The pattern {nm_pattern} should NOT be matched.')
```

hfudev (changes requested):
LGTM! left some minor comments

pytest-embedded/pytest_embedded/dut.py
horw: For me, isinstance(re.compile(b'hello').pattern, str) returns False. So, I think it will work as expected. This line is actually used to convert a string pattern to a byte-pattern.

hfudev: oh, yes. Please ignore this one :D

pytest-embedded/pytest_embedded/dut.py
hfudev: forget about this nitpick. improve next time

PR #260, opened 2024-01-11
fix: use socket bind to get allowed port

horw:
While running the test, we encountered a conflict with port allocation. The current PR is an attempt to resolve this issue.

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -39,7 +39,7 @@ jobs:
           apt update && apt install -y socat zip
           pip install -U pip
           export PIP_EXTRA_INDEX_URL="https://dl.espressif.com/pypi"
-          pip install cryptography --only-binary :all:
+          pip install cryptography --prefer-binary
           pip install -r requirements.txt
           bash foreach.sh install
       - name: Check ports
--- a/pytest-embedded-qemu/pytest_embedded_qemu/qemu.py
+++ b/pytest-embedded-qemu/pytest_embedded_qemu/qemu.py
@@ -1,6 +1,7 @@
 import asyncio
 import os
 import shlex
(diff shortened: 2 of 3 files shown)


horw:
One of the solutions I think we can use is port 0.
There are a few ways to implement this:
- Use a socket and assign port 0, then release it for use in another process.
- Another solution is to make QEMU allocate a port in the format 127.0.0.1:0. However, to obtain this port, we need to install an additional package, such as psutil.
So I chose the first solution, as I don't believe there is a very high probability of a port allocation error.

hfudev:
Thank you. LGTM. Let's fix the ci problems first.

horw:
> Thank you. LGTM. Let's fix the ci problems first.

I have taken a look at this problem:
- Our architecture is armv7l, and the Python version is 3.7.
- The newest version is not compatible with our environment.
- So, I have decided to use a fixed version and set it to 40.0.1.

.github/workflows/test-python.yml
```diff
@@ -39,7 +39,11 @@ jobs:
           apt update && apt install -y socat zip
           pip install -U pip
           export PIP_EXTRA_INDEX_URL="https://dl.espressif.com/pypi"
-          pip install cryptography --only-binary :all:
+          if ${{ matrix.arch  == 'arm' }}; then
+            pip install cryptography==40.0.1
+          else
+            pip install cryptography --only-binary :all:
+          fi
```
hfudev, line 46: try `--prefer-binary`

>   --prefer-binary             Prefer binary packages over source packages, even if the source packages are newer.

horw: Cool! This works

hfudev:
Thank you!

Issue #233, opened 2023-10-24
Wokwi Featured Functions


hfudev:
Wokwi support more functionalities other than serial I/O. like pushing buttons, set/get the state of sensors (described [here](https://docs.wokwi.com/wokwi-ci/getting-started))

Once the API is settled, implement this in pytest-embedded-wokwi

hayschan:
Hello Hanxi,

Thank you for your response in the [previous discussion](https://github.com/hfudev/esp-idf-build-and-test-ci-template/issues/2) regarding the CI/CD docker.

Over the past few days, I've been delving into the Unity unit testing framework and `pytest-embedded` as part of our ESP-IDF software development. I'm currently working on a project that involves testing a matrix keyboard, and I've successfully completed the development of the `test_app`. To streamline the testing process and eliminate the need for manual intervention (i.e., physically pressing the keyboard), I am considering the implementation of Wokwi.

However, I've encountered a couple of areas where additional resources or documentation would be immensely helpful:
1. API Documentation: The [API section for pytest-embedded on Wokwi](https://docs.espressif.com/projects/pytest-embedded/en/latest/api.html#pytest-embedded-wokwi) appears to be incomplete or unpopulated at this time.
2. Practical Examples: The current repository lacks examples that demonstrate the integration of Wokwi with `pytest-embedded`. 

Furthermore, the [existing examples](https://github.com/espressif/pytest-embedded/tree/main/examples/esp-idf), such as confirming a "Hello World" output, don't fully represent complex, real-world scenarios. It would be highly advantageous to have examples that cover more intricate use cases, like I2C interaction or WiFi connectivity testing, timers.

I believe addressing these areas would significantly enhance the utility and applicability of `pytest-embedded` for more complex development environments.

Looking forward to further updates or guidance on these matters.

Best regards.

hfudev:
Hi @hayschan! Thank you for your input.

> API Documentation

`pytest-embedded-wokwi` is a wrapper of the [wokwi-cli](https://docs.wokwi.com/wokwi-ci/getting-started#cli-usage). For now both `wokwi-cli` and `pytest-embedded-wokwi` are in alpha. We're working together with @urish to expose more APIs for testing.

> Practical Examples

[...]

PR #261, opened 2024-01-16
fix(wokwi): run in interactive mode

urish:
see https://github.com/espressif/pytest-embedded/issues/233#issuecomment-1893741907 for details.

Note: this fix requires updating wokwi-cli to the latest version, 0.9.0. For people using this in CI, the latest version should be installed automatically.


hfudev:
Thank you for the quick improvement! @urish

PR #264, opened 2024-02-11
feat(wokwi): support for specifying simulation timeout

### LINKED ISSUE #263: Unable to change wokwi timeout. (RDT-680)
trying to use existing qemu tests with wokwi - they take some time to complete so pytest-embedded-wokwi times out (30 sec default)

Can't seem to find any way to increase the timeout. The [wokwi github action](https://github.com/wokwi/wokwi-ci-action) has configurable timeout, but pytest-embedded-wokwi does not.

urish:
fixes #263

### DIFF
--- a/pytest-embedded-wokwi/README.md
+++ b/pytest-embedded-wokwi/README.md
@@ -35,3 +35,17 @@ Windows PowerShell:
 ```powershell
 $env:WOKWI_CLI_TOKEN="your-api-key"
 ```
+
+#### Usage
+
+To run your tests with Wokwi, make sure to specify the `wokwi` service when running pytest, e.g.:
+
+```
+pytest --embedded-services idf,wokwi
+```
+
+To limit the amount of simulation time, use the `--wokwi-timeout` flag. For example, to set the simulation time limit to 60 seconds (60000 milliseconds):
+
+```
+pytest --embedded-services idf,wokwi --wokwi-timeout=60000
+```
--- a/pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
+++ b/pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
@@ -38,6 +38,7 @@ class WokwiCLI(DuplicateStdoutPopen):
         self,
         firmware_resolver: IDFFirmwareResolver,
         wokwi_cli_path: t.Optional[str] = None,
+        wokwi_timeout: t.Optional[int] = None,
         app: t.Optional['IdfApp'] = None,
         **kwargs,
     ):
@@ -52,9 +53,12 @@ class WokwiCLI(DuplicateStdoutPopen):
         self.create_diagram_json()
 
         wokwi_cli = wokwi_cli_path or self.wokwi_cli_executable
+        cmd = [wokwi_cli, '--interactive', app.app_path]
+        if (wokwi_timeout is not None) and (wokwi_timeout > 0):
+            cmd.extend(['--timeout', str(wokwi_timeout)])
 
         super().__init__(
-            cmd=[wokwi_cli, '--interactive', app.app_path],
+            cmd=cmd,
             **kwargs,
         )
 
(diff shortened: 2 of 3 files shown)


pytest-embedded-wokwi/README.md
```diff
@@ -35,3 +35,17 @@ Windows PowerShell:
...
+To run your tests with Wokwi, make sure to specify the `wokwi` service when running pytest, e.g.:
+
+```
+pytest --embedded-services idf,wokwi
+```
+
+If your tests run for longer than 30 seconds, you may have to increase the simulation timeout. For example, to set the timeout to 60 seconds (60000 milliseconds):
+
+```
+pytest --embedded-services idf,wokwi --wokwi-timeout=60000
```
igrr, line 50: Shouldn't this somehow be derived from the timeouts specified in dut.expect(...) calls? Otherwise, we have to specify the timeout in two places: in dut.expect and as a CLI argument.

Another option, can we simply let the simulation run without a timeout? Timeout in dut.expect will still interrupt the simulation if the expressions isn't matched, which will fail the test and stop the simulation.

urish: Thanks for the feedback!

Since the `dut.expect(...)` calls are only evaluated at runtime, we can't know in advance how long the timeout should be.

The purpose of the timeout is to safeguard against errors (such as a case where the test gets stuck in an infinite loop). The simulation is computationally expensive, so we wouldn't want this process to keep running for hours in case of such error. We can bump the default from 30 seconds to a larger value, but from my experience so far, 30 seconds is enough for more test cases (that's why it's also the current default).

igrr: Right, but why would the simulation be running for such a long time without the pytest test case exiting?

If the test case calls `dut.expect(...)` and the simulation doesn't output the expected expression, the test case will fail and the simulation will get stopped.

urish: Normally it wouldn't. For wokwi-cli (in general), it makes sense to have a time out since it's very easy to just run it in a terminal and forget it running. 

In the specific case of pytest embedded, I think only some sort of developer error could cause it to run for a longer time (e.g. users trying to implement their own expect mechanism instead of using `dut.expect()`. 

At some point we plan to start billing for CI usage, so, what I'm essentially worried about is the user experience - if a developer had such error and the CI keeps running for hours while it would normally complete within seconds, it could become rather expensive for the user.

hfudev: @urish I totally agree that setting a default timeout could be helpful for developers cutting unneeded expenses. However, I'd like to propose adding a complementary flag, `--long-lasting`, to explicitly disable the timeout for certain use cases.

This flag will mostly benefit users who integrate wokwi-cli into CI platforms like GitHub Actions or GitLab. These platforms often have their own timeout settings, and providing the `--long-lasting` flag would allow users to configure the timeout centrally while still respecting Wokwi's cost considerations.

I believe this approach balances both cost concerns and user flexibility. WDYT?

urish: Let's see if I get it right, the suggestion is to add a `--long-lasting` flag to the Wokwi CLI, and then use it always when running from within pytest-embedded? Or do we still want to keep the `--wokwi-timeout` flag for pytest and use `--long-lasting` as the default if no timeout provided?

hfudev: I'd prefer the latter one, keep the `--wokwi-timeout` flag and use `--long-lasting` as the default if no timeout specified. Thanks!

urish: Alright, I pushed a change where I kept `--wokwi-timeout`, and set the default to one day (which is in essence the same as `--long-lasting`). I also updated the documentation to clarify that this new flag is limiting the simulation time.

hfudev: sure. thanks for the update!

PR #266, opened 2024-02-29
fix(idf): check sdkconfig items while `erase_flash`

hfudev:


horw (commented):
LGTM! Thanks for these changes.

PR #272, opened 2024-04-02
fix: deprecate `use_esptool` args correctly

hfudev:


horw:
Cool, this way for deprecating errors looks much better.

PR #273, opened 2024-04-02
fix: flash baud

### LINKED ISSUE #268: Esptool Flash is Slow (RDT-722)
Now we're using esptool.main flash, with baudrate 115200.

For flashing baudrate doesn't matter at all, we could set to 1,000,000 to make the flash faster.

horw:
Set the baud rate to 1,000,000 to make the flash faster when no value is set up. 
Closes #268.

### DIFF
--- a/pytest-embedded-idf/pytest_embedded_idf/serial.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/serial.py
@@ -1,6 +1,7 @@
 import contextlib
 import hashlib
 import logging
+import os
 import tempfile
 from typing import Optional, TextIO, Union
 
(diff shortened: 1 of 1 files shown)


pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -161,6 +162,8 @@ def flash(self, app: Optional[IdfApp] = None) -> None:
                 else:
                     _args.append(str(v))
 
+        if '--baud' not in _args:
+            _args.extend(['--baud', os.getenv('ESPBAUD', '1000000')])
         _args.append('write_flash')
```
hfudev, line 167: I'll merge it as is. next time better to add a blank line here, to make code blocks referring different logic separated.

```suggestion
            _args.extend(['--baud', os.getenv('ESPBAUD', '1000000')])
            
        _args.append('write_flash')
```

hfudev:
LGTM!

PR #277, opened 2024-04-25
feature: Support Arduino in Wokwi

P-R-O-C-H-Y:
Adds support for pytest-embedded-wokwi to run with Arduino build files.


pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
```diff
@@ -101,9 +102,16 @@ def create_wokwi_toml(self):
 
     def create_diagram_json(self):
         app = self.app
-        diagram_json_path = os.path.join(app.app_path, 'diagram.json')
         target_board = target_to_board[app.target]
 
+        # Check for specific target.diagram.json file first
+        diagram_json_path = os.path.join(app.app_path, (app.target + '.diagram.json'))
+        if os.path.exists(diagram_json_path):
+            shutil.copyfile(diagram_json_path, os.path.join(app.app_path, 'diagram.json'))
```
hfudev, line 110: here we're copying the `esp32.diagram.json` to `diagram.json`. what will happen if user has both files? the `diagram.json` will be overwritten.

does wokwi support setting the diagram path?
- if yes, we could use `wokwi --diagram-filepath esp32.diagram.json` directly
- if no, we should backup user's `diagram.json` to `diagram.json.old` first (when exist), then do the copy

hfudev (changes requested):

pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
P-R-O-C-H-Y: I will check and update accordingly.

pytest-embedded-wokwi/tests/test_wokwi.py
```diff
@@ -36,3 +36,22 @@ def test_pexpect_by_wokwi(dut):
     )
 
     result.assert_outcomes(passed=1)
+
+
+def test_pexpect_by_wokwi_esp32_arduino(testdir):
```
hfudev, line 41:
```suggestion
@wokwi_cli_required
@wokwi_token_required
def test_pexpect_by_wokwi_esp32_arduino(testdir):
```

hfudev (changes requested):

pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
P-R-O-C-H-Y: There is no argument to specify diagram.json file path.
So I will backup the diagram.json to diagram.json.old. Do we need to restore it afterwards or check also for the .old one?

pytest-embedded-wokwi/tests/test_wokwi.py
P-R-O-C-H-Y: Done

pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
P-R-O-C-H-Y: I think the main usage will be just single diagram.json or multiple of them specifying it for each target.
I dont think both options together will be used, it doesn't make much sense.

pytest-embedded-wokwi/pytest_embedded_wokwi/wokwi_cli.py
```diff
@@ -101,9 +102,23 @@ def create_wokwi_toml(self):
...
         app = self.app
-        diagram_json_path = os.path.join(app.app_path, 'diagram.json')
         target_board = target_to_board[app.target]
 
+        # Check for specific target.diagram.json file first
+        diagram_json_path = os.path.join(app.app_path, (app.target + '.diagram.json'))
+        if os.path.exists(diagram_json_path):
+            # If there is also common diagram.json file, backup it first to diagram.json.old
+            if os.path.exists(os.path.join(app.app_path, 'diagram.json')):
+                shutil.copyfile(
```
hfudev, line 112: let's give user some warning here

```python
logging.warning('using %s instead. backup the original diagram.json to diagram.json.old', diagra_json_path)
```

P-R-O-C-H-Y: Done :)

hfudev (approved):

hfudev:
Thank you for this PR!

PR #282, opened 2024-04-26
feat: flexible choices for run_all_single_board_cases

horw:
The Unity test framework itself supports different inputs for choice tests. This merge request emulates this behavior to make it possible to control test choices with pytest-embedded.

The current ways to choose tests are:
- group
- invert group
- name

For example:
```
2024-04-26 16:35:17 Here's the test menu, pick your combo:
2024-04-26 16:35:17 (1)	"can use static initializers for non-POD types" [restart_init]
2024-04-26 16:35:17 (2)	"static initialization guards work as expected" [misc]
2024-04-26 16:35:17 (3)	"global initializers run in the correct order" [misc]
2024-04-26 16:35:17 (4)	"before scheduler has started, static initializers work correctly" [misc]
2024-04-26 16:35:17 (5)	"init_priority extension works" [misc]
2024-04-26 16:35:17 (6)	"can use new and delete" [misc]
2024-04-26 16:35:17 (7)	"can call virtual functions" [misc]
2024-04-26 16:35:17 (8)	"can use std::vector" [misc]
2024-04-26 16:35:17 (9)	"stack smashing protection CXX" [stack_smash]
```
1. choose all (1-9):
    dut.run_all_single_board_cases()
2. choose by group(2-8):
    dut.run_all_single_board_cases(group="misc")
3. choose by inverting group(1,9):
    dut.run_all_single_board_cases(group="!misc")
4. choose by name(9):
    dut.run_all_single_board_cases(names=["stack smashing protection CXX"])


horw:
There are some unit test cases that use multiple groups and arguments to filter test cases, for example: [heap][mem_prot] and [test_env=noXtal32k]. So, I have added these options.

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -422,23 +422,47 @@ def run_all_single_board_cases(
...
     ) -> None:
         """
         Run all single board cases, including multi_stage cases, and normal cases
 
         Args:
-            group: test case group
-            reset: whether to perform a hardware reset before running a case
-            timeout: timeout. (Default: 30 seconds)
-            run_ignore_cases: run ignored test cases or not
+            group: Test case group. Supports group invert by '!' and groups separated by ','.
```
hfudev, line 432: hmm, what will happen when I write '!a,b', it will be explained as which one?

- not in group a, or in group b
- not in group a or b

improve the docstring with `.. note:` would be nice.

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -422,23 +422,47 @@ def run_all_single_board_cases(
...
+            reset: Whether to perform a hardware reset before running a test case.
+            timeout: Timeout duration in seconds. Default is 30 seconds.
```
hfudev, line 434:
```suggestion
            timeout: Timeout in seconds. (Default: 30 seconds)
```

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
hfudev: besides, groups separated by `,` is not a nice idea, since there's no such limitation in our c code. if there's a group named 'a,b', then we got different behavior.

I suggest we support:

```
group: Test case group or a list of test case groups.
```

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -422,23 +422,47 @@ def run_all_single_board_cases(
...
+            run_ignore_cases: Whether to run ignored test cases or not.
+            names: List of test case names to run.
```
hfudev, line 436:
```suggestion
            name: test case name or a list of test case names to run.
```

same as `group`, better support both str and list[str]

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -422,23 +422,47 @@ def run_all_single_board_cases(
...
+            attributes: Dictionary of attributes to filter and run test cases.
         """
+        inverted = group and group.startswith('!')
```
hfudev, line 439: if we support ['a', '!b'] then the current logic needs to be changed.

hfudev (changes requested):
some comments regarding api, overall the code LGTM.

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
horw: Should we support this behavior? :eyes:

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
horw: Yes, I agree, this is not a good idea. Let's make them a str or list then.

horw: > * not in group a, or in group b
> * not in group a or b

For now it will `not (a and b)`

hfudev: if the logic operand between each list item is "and" it's clear, but counter-intuitive...

For example,

- ['!a', '!b'] is: not in group a, and not in group b
- ['!a', 'b'] is: not in group a, and in group b

if the user want to run group a and group b, they need to call the function twice...

horw: Let's do it using 'and' because the unit test for [a][b] will be done using 'and' and not 'or'.
I have created function `validate_group` for it

horw: On the other hand, it is easier to just `run_all_single_board_cases` using this function a few times than to create a more complex interface for it.

hfudev: > Let's do it using 'and' because if we compare it to the unit test for [a][b], it will be done using 'and' and not 'or'.

actually on the contrary... unity test case `"test foo" [a][b]` means it could be selected both by `a`, or by `b`.

horw: Let's do it like "a & b | c"
WDYT

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -418,27 +418,70 @@ def run_single_board_case(self, name: str, reset: bool = False, timeout: float =
...
-                        self._run_multi_stage_case(case, reset=reset, timeout=timeout)
+            should_run = False
+            if not group and not name and not attributes:
+                should_run = True
+            if group and validate_group(case.groups):
+                should_run = True
+            if name and case.name in name:
+                should_run = True
+            if attributes and all(case.attributes.get(k) == v for k, v in attributes.items()):
+                should_run = True
```
hfudev, line 476:
```suggestion
            if group and validate_group(case.groups):
                should_run = True
            if name and case.name in name:
                should_run = True
            if attributes and all(case.attributes.get(k) == v for k, v in attributes.items()):
                should_run = True
```

here the logic is not clear in the documentation.

if the user specified `group`, `name`, and `attributes` all together, test cases that matched "group" or "name" or "attribute" would be run.

I think the behavior is okay, but we could put a `Notes` in the docstring.

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -418,27 +418,70 @@ def run_single_board_case(self, name: str, reset: bool = False, timeout: float =
...
+            return False
+
         for case in self.test_menu:
-            if not group or group in case.groups:
-                if not case.is_ignored or run_ignore_cases:
-                    if case.type == 'normal':
-                        self._run_normal_case(case, reset=reset, timeout=timeout)
-                    elif case.type == 'multi_stage':
-                        self._run_multi_stage_case(case, reset=reset, timeout=timeout)
+            should_run = False
```
hfudev, line 468: the `should_run` argument is not accurate. `ignore` will be considered later.

here better to rename to `selected`

pytest-embedded-idf/tests/test_idf.py
```diff
@@ -716,6 +716,90 @@ def test_dut_run_all_single_board_cases(dut):
...
+def test_dut_run_all_single_board_cases_group(testdir):
+    testdir.makepyfile(r"""
+        def test_dut_run_all_single_board_cases(dut):
+            dut.run_all_single_board_cases(group="normal_case", timeout=10)
+    """)
+    testdir.runpytest(
+        '-s',
+        '--embedded-services', 'esp,idf',
+        '--app-path', os.path.join(testdir.tmpdir, 'unit_test_app_esp32'),
+        '--log-cli-level', 'DEBUG',
```
hfudev, line 728: you're not checking the debug level in the test case, maybe remove the log cli level here?

pytest-embedded-idf/tests/test_idf.py
```diff
@@ -716,6 +716,90 @@ def test_dut_run_all_single_board_cases(dut):
...
+
+    junit_report = ET.parse('report.xml').getroot()[0]
+
+    assert junit_report.attrib['errors'] == '0'
+    assert junit_report.attrib['failures'] == '1'
+    assert junit_report.attrib['skipped'] == '0'
+    assert junit_report.attrib['tests'] == '1'
+
+
+def test_dut_run_all_single_board_cases_or_groups(testdir):
```
hfudev, line 740: remove this test?

hfudev (changes requested):
LGTM!

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
horw: Yes there is 'or' condition. I will note it

pytest-embedded-idf/tests/test_idf.py
horw: yes, sure

pytest-embedded-idf/tests/test_idf.py
horw: removed

PR #285, opened 2024-04-29
fix(qemu): stop require optional dependency 'idf'

hfudev:


pytest-embedded-qemu/pytest_embedded_qemu/__init__.py
```diff
@@ -1,12 +1,28 @@
...
 
+
+__getattr__ = lazy_load(
+    importlib.import_module(__name__),
+    {
+        'Qemu': Qemu,
+        'QemuDut': QemuDut,
+    },
+    {
+        'QemuApp': '.app',
```
hfudev, line 21: maybe rename to `QemuIdfApp` in the future...

PR #286, opened 2024-04-30
fix(wokwi): Fix paths for bin and elf files

P-R-O-C-H-Y:
Fixes wrong paths (or unsupported paths) for binary and elf files for Wokwi.


hfudev:
could you help update the commit message?

`Fix paths for bin and elf files` is not accurate.

`error when bin,elf file not under app_path` would be more helpful.

PR #289, opened 2024-05-06
feat: dry run single board unity test

horw:
This MR implements a feature for dry running single board Unity tests, allowing users to display a list of test case names without actually running them when the `dry_run` parameter in `run_all_single_board_cases` is set to True.

### DIFF
--- a/pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
@@ -416,6 +416,34 @@ class IdfUnityDutMixin:
         else:
             raise ValueError(f'single-board test case {name} not found')
 
+    @staticmethod
+    def _select_to_run(group, name, attributes, case_groups, case_name, case_attributes):
+        def validate_group():
+            for _or in group:
+                for _and in _or:
+                    invert = _and.startswith('!')
+                    _and = _and.lstrip('!')
+                    result = _and in case_groups
+                    if invert:
+                        result = not result
+                    if not result:
+                        break
+                else:
+                    return True
+
+            return False
+
+        if not group and not name and not attributes:
+            return True
+        if group and validate_group():
+            return True
+        if name and case_name in name:
+            return True
+        if attributes and all(case_attributes.get(k) == v for k, v in attributes.items()):
+            return True
+
+        return False
+
     def run_all_single_board_cases(
         self,
         group: t.Optional[t.Union[str, list]] = None,
@@ -424,6 +452,7 @@ class IdfUnityDutMixin:
         run_ignore_cases: bool = False,
         name: t.Optional[t.Union[str, list]] = None,
         attributes: t.Optional[dict] = None,
+        dry_run: bool = False,
     ) -> None:
         """
         Run all single board cases, including multi_stage cases, and normal cases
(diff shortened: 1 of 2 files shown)


pytest-embedded-idf/tests/test_idf.py
```diff
@@ -682,6 +682,70 @@ def test_idf_hard_reset_and_expect(dut):
     result.assert_outcomes(passed=1)
 
 
+def test_select_to_run():
+    from pytest_embedded_idf.unity_tester import IdfUnityDutMixin
+
+    assert IdfUnityDutMixin._IdfUnityDutMixin__select_to_run(
```
hfudev, line 688:
```suggestion
    assert IdfUnityDutMixin._select_to_run(
```

let's name it `_select_to_run`, and update the usage like this...

the `__` is somehow not recommended in python, since it refers to "magic functions"

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -450,37 +480,15 @@ def run_all_single_board_cases(
...
-            if name and case.name in name:
-                selected = True
-            if attributes and all(case.attributes.get(k) == v for k, v in attributes.items()):
-                selected = True
+            selected = self.__select_to_run(group, name, attributes, case.groups, case.name, case.attributes)
 
             if not selected:
                 continue
+            if dry_run:
+                logging.info('%s | %s', case.index, case.name)
```
hfudev, line 489: could you make it happen under the "ignore" block later?

this could be confusing when one test case is "[ignored]", showing here, but not been run later.

hfudev (changes requested):
LGTM!

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -440,6 +469,7 @@ def run_all_single_board_cases(
             run_ignore_cases: Whether to run ignored test cases or not.
             name: test case name or a list of test case names to run.
             attributes: Dictionary of attributes to filter and run test cases.
+            dry_run: If True, then just show a list of test case names without running them.
```
hfudev, line 472:
```suggestion
            dry_run: If True, then just show a list of test case names without running them. (please set the logging level as INFO to see them)
```

otherwise the default `WARNING` level won't show these `logging.info` lines

pytest-embedded-idf/pytest_embedded_idf/unity_tester.py
```diff
@@ -450,37 +481,15 @@ def run_all_single_board_cases(
...
-                selected = True
+            selected = self._select_to_run(group, name, attributes, case.groups, case.name, case.attributes)
 
             if not selected:
                 continue
+            if dry_run:
+                logging.info('[ignored] %s | %s', case.index, case.name)
+                continue
+
             if not case.is_ignored or run_ignore_cases:
```
hfudev, line 493:
```suggestion
            if not case.is_ignored or run_ignore_cases:
                if dry_run:
                    logging.info('[dry run] %s | %s', case.index, case.name)
                    continue
```

hfudev (changes requested):

hfudev:
LGTM, only minor issues, don't forget to improve the user experience in your future PRs...

PR #298, opened 2024-07-26
fix/flash port

horw:
In this MR, I have changed logic for flash port


pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -408,6 +412,21 @@ def serial_gn(_fixture_classes_and_options, msg_queue, app) -> t.Optional[t.Unio
...
+
+    if kwargs.get('flash_port'):
+        operation_port = kwargs.pop('port', None)
+        if operation_port is None:
+            raise SystemExit('If the flash port was set up, the port should also be set up.')
+
+        flash_port = kwargs.pop('flash_port')
+        kwargs['stop_after_init'] = True
+        kwargs['port'] = flash_port
+        cls(**_drop_none_kwargs(kwargs))
```
hfudev, line 424: this instance has not been returned, so the teardown won't happen in `multi_dut_generator_fixture`.

in `multi_dut_generator_fixture`, we only teardown the obj that returns.

```python
...
            try:
                res = func(*args, **kwargs)
                yield res
            finally:
                if res:
                    _close_or_terminate(res)
```

horw: @hfudev Hello, you are right about multiple port occupation for dut's. Therefore, I have updated the occupation logic. Please take a look.

hfudev:
The new approach LGTM!

pytest-embedded-serial-esp/pytest_embedded_serial_esp/serial.py
```diff
@@ -67,8 +67,9 @@ def __init__(
         self._meta = meta
 
         esptool_target = beta_target or target or 'auto'
-        if port is None:
-            available_ports = esptool.get_port_list()
+        if port is None or port.endswith('*'):
+            port_filter = port.strip('*') if port else ''
+            available_ports = [_p for _p in esptool.get_port_list() if port_filter in _p]
```
hfudev, line 72: here should filter out the `ports_to_occupy` as well, since they're "released" by the first-dut's `close`

pytest-embedded-serial/pytest_embedded_serial/serial.py
```diff
@@ -47,13 +47,16 @@ def __init__(
         port_location: Optional[str] = None,
         baud: int = DEFAULT_BAUDRATE,
         meta: Optional[Meta] = None,
+        stop_after_init: bool = False,
+        ports_to_occupy: List[str] = (),
```
hfudev, line 51: should be put in all the subclasses to filter these ports out while auto-detecting...

horw: Sure, thank you!

horw: @hfudev  I have added it to EspSerial. ArduinoDut and IdfDut don't need it.

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -264,7 +264,7 @@ def setup_jtag(self):
 
     def flash_via_jtag(self):
         if not self.openocd:
-            logging.debug('no openocd instance created. NOT flashing via openocd `program_esp`')
+            logging.warning("no openocd instance created. can't flash via openocd `program_esp`")
```
hfudev, line 267: debug

horw: Solved!

hfudev:
Thank you! LGTM!

PR #299, opened 2024-08-07
fix(jtag): set gdb remote timeout to 10s by default

hfudev:


erhankur:
@hfudev LGTM. Thanks for the quick PR.

PR #300, opened 2024-08-09
fix(jtag): gdb write "remotetimeout" stop waiting for response

hfudev:


erhankur:
Hopefully, this will avoid no response timeout. LGTM. thanks @hfudev

PR #302, opened 2024-08-10
fix(jtag): flush previous gdb responses before the first gdb command (RDT-857)

erhankur:
In many test logs, I see a 'gdb no response' error. As I understand, this occurs due to a loss of synchronization between what we actually send and what is inside the response file. With this PR, we will ensure that before the first command, there will be no unevaluated response in the queue.

I also made some small changes that seem more logical to me.

### DIFF
--- a/pytest-embedded-jtag/pytest_embedded_jtag/gdb.py
+++ b/pytest-embedded-jtag/pytest_embedded_jtag/gdb.py
@@ -20,35 +20,38 @@ class Gdb(DuplicateStdoutPopen):
         gdb_prog_path = gdb_prog_path or self.GDB_PROG_PATH
         gdb_cli_args = shlex.split(gdb_cli_args or self.GDB_DEFAULT_ARGS)
 
-        self._gdb_first_prompt_matched = False
+        self._gdb_first_write = True
 
         super().__init__(cmd=[gdb_prog_path, *gdb_cli_args], **kwargs)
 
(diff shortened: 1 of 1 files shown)


erhankur:
@hfudev PTAL

pytest-embedded-jtag/pytest_embedded_jtag/gdb.py
```diff
@@ -20,36 +20,39 @@ def __init__(self, gdb_prog_path: Optional[str] = None, gdb_cli_args: Optional[s
...
+                # Discard all queued responses before the first write
+                _ = fr.readlines()
+                self._logfile_offset = fr.tell()
+                self._gdb_first_write = False
+
+            super().write(s)
+
+            if non_blocking:
+                logging.debug('non-blocking write...')
+                return None
```
hfudev, line 39: I feels like we still need to readlines after the non-blocking write.

otherwise if one non-blocking write generates the `(gdb)`, it could affect the expect happening later in a second blocing write

```
# non-blocking write
(gdb)
# blocking write happend, it would match the (gdb) returned above
# maybe still pending here
```

hfudev: Yes the usage in the test scripts is correct. but I think there's a potential bug existing before this PR.

The problem is that, if we're writing a command that could return `(gdb)` with non-blocking, the latter blocking write may have issue since it could expect on the `(gdb)` returned by the earlier non-blocking command.

We could fix it in a separate PR.

PR #311, opened 2024-10-09
fix: close serial port when exception

### LINKED ISSUE #307: Serial File Descriptor Not Closed When Error Occurs in Serial.__init__ (RDT-912)
**Describe the bug**
```python
  res = None
  try:
      res = func(*args, **kwargs)
      yield res
  finally:
      if res:
          _close_or_terminate(res)
```
This part of the code only considers `close_or_terminate` when the object has already been created. However, if the object gets stuck in `__init__`, the serial file descriptor, for example, will never be closed. A similar bug may occur with other objects as well.

**To Reproduce**
1. Raise any error in the `self._start()` function of the `Serial` class.
2. `self.proc` will never be closed.

**Expected behavior**
All objects should be processed properly. If there is an exception, any opened resources should be closed.

horw:
## Description
Fixing issue with not closed serial

## Related

Closes #307

## Testing

---

## Checklist


pytest-embedded-serial/pytest_embedded_serial/serial.py
```diff
@@ -90,14 +90,18 @@ def __init__(
             self.proc = pyserial.serial_for_url(self.port, **port_config)
 
         self.ports_to_occupy.append(self.port)
-        self._post_init()
-        self._start()
+        try:
+            self._post_init()
+            self._start()
```
hfudev, line 95: try catch only this function?

horw: sure

PR #314, opened 2024-10-24
fix(wokwi): Use merged bin to ensure partition and bootloader flashing

lucasssvaz:
## Description

Arduino tests in Wokwi will fallback to a default bootloader and partition table if none is provided. This can cause issues with some tests. To make sure the partition table and bootloader are always flashed, we need to use the generated merged image by the Arduino compilation.

## Testing

Tested locally, works as expected.

### DIFF
--- a/pytest-embedded-wokwi/pytest_embedded_wokwi/arduino.py
+++ b/pytest-embedded-wokwi/pytest_embedded_wokwi/arduino.py
@@ -12,4 +12,4 @@ class ArduinoFirmwareResolver:
 
     def resolve_firmware(self, app: 'ArduinoApp'):
         # get path of ino.bin file
-        return Path(app.binary_path, app.sketch + '.ino.bin')
+        return Path(app.binary_path, app.sketch + '.ino.merged.bin')


pytest-embedded-wokwi/pytest_embedded_wokwi/arduino.py
```diff
@@ -12,4 +12,4 @@ class ArduinoFirmwareResolver:
 
     def resolve_firmware(self, app: 'ArduinoApp'):
         # get path of ino.bin file
-        return Path(app.binary_path, app.sketch + '.ino.bin')
+        return Path(app.binary_path, app.sketch + '.ino.merged.bin')
```
hfudev, line 15: One question, is this file auto generated by arduino? if no we can't do the rename here I suppose...

besides, the naming in `wokwi_cli.py` looks a bit weird...

```
        flasher_args = self.firmware_resolver.resolve_firmware(app) # <-- should be firemware?
        wokwi_toml_path = os.path.join(app.app_path, 'wokwi.toml')
        firmware_path = Path(os.path.relpath(flasher_args, app.app_path)).as_posix()
        elf_path = Path(os.path.relpath(app.elf_file, app.app_path)).as_posix()
```

could you help update it as well?

lucasssvaz: Yes, this file is auto generated. So we can change use it consistently.

About the naming I can take a look but probably @urish knows if it is correct or not.

hfudev: no problem. thanks.

PR #317, opened 2024-11-07
feat(wokwi): Use new board for ESP32-P4 (RDT-1002)

lucasssvaz:
## Description

Change the default P4 board on Wokwi from preview to the latest EV board.

## Testing

Tested locally


hfudev:
Thank you @lucasssvaz !

PR #318, opened 2024-11-11
Add support for NuttX RTOS, with optional support for Espressif devices (RDT-1004)

fdcavalcanti:
## Description
This PR adds support for NuttX RTOS and Espressif devices.
- DUT: new DUT class for NuttX, supporting write and return, return codes and reboot.
- Serial: Supports flashing Espressif devices (ESP32|S2|S3|C3|C6|H2).
- App: Supports parsing binaries for Espressif device (both app and bootloader), extracting flash information automatically, which is used by the serial class to flash.

Looking forward for reviews and improvement suggestions!

## Testing
Tested on Espressif internal CI for NuttX. The entire test suite has been ported to Pytest.
Local test environment for this tool simply follows common pip instructions already available in the docs.

---

## Checklist


pytest-embedded-nuttx/pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python",
+    "Topic :: Software Development :: Testing",
+]
+dynamic = ["version", "description"]
+requires-python = ">=3.7"
+
+dependencies = [
+    "pytest-embedded-serial-esp~=1.11.8",
```
hfudev, line 34: I understand the `service` is optional

- `esp,nuttx` -> `NuttxEspDut`, `NuttxApp`, `NuttxSerial`
- `nuttx` -> `NuttxDut`, `Serial`

however, here it's added as a hard-dependency, You may place `pytest-embedded-serial` under `[project.dependencies`, and `pytest-embedded-serial-esp` under [project.optional-dependencies]? https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements

pytest-embedded-nuttx/pytest_embedded_nuttx/app.py
```diff
@@ -0,0 +1,100 @@
...
+        """Binary target should be in the format nuttx.merged.bin, where
+        the 'merged.bin' extension can be modified by the file_extension
+        argument.
+
+        Important note regarding MCUBoot:
+        If enabled, the magic number will be on the MCUBoot binary. In this
+        case, image_info should run on the mcuboot binary, not the NuttX one.
+        """
+
+        def get_key_from_value(dictionary, val):
```
hfudev, line 77: I'm unaware of the possible inputs, so just one question

would it be possible to have two keys with same value?

the current behavior would return the first match value, shall we raise an exception when there are two same value keys?

pytest-embedded-nuttx/pytest_embedded_nuttx/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    importlib.import_module(__name__),
+    {
+        'NuttxApp': NuttxApp,
+    },
+    {
+        'NuttxSerial': '.serial',  # requires 'esp'
+        'NuttxEspDut': '.dut',  # requires 'esp'
+        'NuttxDut': '.dut',  # requires 'serial'
+    },
+)
```
hfudev, line 19: `NuttxDut`, and `NuttxEspDut` shall be imported normally,

(although `NuttxEspDut` is activated with service `esp`, but it could be imported normally since it does not require extra dependencies)

`NuttxApp`, and `NuttxSerial` requires esptool / pytest-embedded-serial-esp, so they shall be lazy-loaded.

pytest-embedded-nuttx/pytest_embedded_nuttx/dut.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Returns:
+            AnyStr
+        """
+        self.write(data)
+        ans = self.expect(pexpect.TIMEOUT, timeout=timeout)
+        return ans.rstrip().decode()
+
+
+class NuttxEspDut(NuttxDut):
```
hfudev, line 78: could also be placed here together with NuttxDut, since it does not require extra dependencies.

hfudev (commented):
LGTM! I left some questions regarding the dependencies.

pytest-embedded-nuttx/pyproject.toml
fdcavalcanti: Yes, that makes more sense. Thanks!

pytest-embedded-nuttx/pytest_embedded_nuttx/app.py
fdcavalcanti: This would not be an issue.
I checked the possible values on the dictionaries used, they do not have two keys with the same value and those won't be changing anytime, since they are very tightly coupled to flash memory stuff.

pytest-embedded-nuttx/pytest_embedded_nuttx/__init__.py
fdcavalcanti: Please check if ok.

pytest-embedded-nuttx/pytest_embedded_nuttx/dut.py
fdcavalcanti: Problem is the `reset_to_nsh` method. A generic device from a different manufacturer won't have the `serial.hard_reset` call, which comes from the esp service.
Keeping it separated like this would allow any device running `nuttx,serial` services to use the NuttxDut.

hfudev: yes. you're right. looks good.

pytest-embedded-nuttx/pytest_embedded_nuttx/app.py
hfudev: sure. no problem

pytest-embedded-nuttx/pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+]
+dynamic = ["version", "description"]
+requires-python = ">=3.7"
+
+dependencies = [
+    "pytest-embedded-serial~=1.11.8",
+]
+
+[project.optional-dependencies]
+serial = [
```
hfudev, line 38: Here I think `esp` is a better name... but since pytest-embedded-arduino is picking `serial` as well, it's also acceptable for keeping consistency.

up to you :)

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -376,6 +392,20 @@ def _fixture_classes_and_options_fn(
                         'skip_check_coredump': skip_check_coredump,
                         'panic_output_decode_script': panic_output_decode_script,
                     })
+                if 'esp' in _services and 'nuttx' in _services:
```
hfudev, line 395:
```suggestion
                elif 'esp' in _services and 'nuttx' in _services:
```

hfudev (commented):
found one more issue :D

pytest-embedded-nuttx/pytest_embedded_nuttx/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+__getattr__ = lazy_load(
+    importlib.import_module(__name__),
+    {
+        'NuttxApp': NuttxApp,
+    },
+    {
+        'NuttxSerial': '.serial',  # requires 'esp'
+    },
+)
```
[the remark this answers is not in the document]
fdcavalcanti: Ah, I see. Got confused on this lazy loading stuff.
Thanks for the fix!!

pytest-embedded-nuttx/pyproject.toml
fdcavalcanti: I agree. Should keep it tidy.

fdcavalcanti:
@hfudev all done! Thanks for the review!

hfudev:
Thank you! Will release today or tomorrow

PR #319, opened 2024-11-13
Fix/optional dependency nuttx

hfudev:


hfudev:
@fdcavalcanti PTAL. I found one more issue with the optional dependency.

fdcavalcanti:
@hfudev looks good.

PR #325, opened 2024-11-29
Fix: nuttx and other small changes

hfudev:


hfudev:
@fdcavalcanti PTAL the nuttx changes. I split the `NuttxQemuDut` to `qemu.py`, since it requires optional dependency `qemu`

fdcavalcanti:
@hfudev looks good to me! Thanks for the heads up.

PR #327, opened 2024-12-23
feat: added skip_if_soc marker for idf target

### LINKED ISSUE #326: [idf] Support Skip Marker for ESP-IDF Soc Caps (RDT-1055)
[Soc Caps](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/soc_caps.html) defines hardware capabilities or ESP-IDF apps.

### 1. Basic

Support a marker, like

```python
@pytest.mark.skip_if_soc("SOC_ADC_SAMPLE_FREQ_THRES_HIGH == 20000")
```

Iterate through all supported targets, if the expression is false, then we skip running the test case for this target.

### 2. Preview Targets

Support iterating through preview targets as well, by 

```python
@pytest.mark.skip_if_soc(
    "SOC_ADC_SAMPLE_FREQ_THRES_HIGH == 20000",
    include_preview_targets=True
)
```

Here we shall iterate over `SUPPORTED_TARGETS + PREVIEW_TARGETS`

### 3. Limited Targets

Or iterate over a subset of targets, if user specified `targets` as param, like

```python
@pytest.mark.skip_if_soc("SOC_ADC_SAMPLE_FREQ_THRES_HIGH == 20000")
@pytest.mark.parametrization('target', ['esp32', 'esp32s2'], indirect=True)
def test_foo():
    pass
```

here shall only iterate over esp32 and esp32s2

## Implementation Details

- the parsing soc caps part could be done by [`idf_build_apps.manifest.if_parser`](https://docs.espressif.com/projects/idf-build-apps/en/latest/references/api/idf_build_apps.manifest.html#idf-build-apps-manifest-package)
- the skip test case part shall be done in the pytest hook function `pytest_collection_modifyitems`
- `idf-build-apps` shall be added as a dependency for `pytest-embedded-idf`
- `SUPPORTED_TARGETS` and `PREVIEW_TARGETS` could be loaded from `idf_build_apps`. These values are lazy loaded from IDF_PATH.
- Make sure everything is lazy-loaded, otherwise it would break user without IDF_PATH.

horw:
## Description
This PR introduces a `skip_if_soc` marker that helps filter targets based on their `SOC_CAPS.h` attributes.

## Related
Closes #326

## Testing

---

## Checklist


pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1253,6 +1254,26 @@ def _duplicate_items(items: t.List[_T]) -> t.List[_T]:
...
+            skip_marker = item.get_closest_marker('skip_if_soc')
+            if not skip_marker:
+                continue
+
+            if not hasattr(item, 'callspec'):
+                logging.warning(
+                    f"Test item '{item.originalname}'. "
+                    f"Ensure 'target' is included in the test @pytest.mark.parametrize when using 'skip_if_soc'."
+                )
+                continue
```
hfudev, line 1268: what about we support `target` from CLI as well (not "auto")?

so it could also be used as `skipif` when user call `pytest --target esp32`

horw: Agree, added

horw:
There is a question about multi-DUT support: should we support it? If yes, I think we need to do something like masking -> `esp32|esp32c5`, where the mask would be `0|1`, or for `esp32c5|esp32c6`, use the default mask(by default check all), or for `esp32c5|esp32`, mask it as`1|0`.

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1252,7 +1253,35 @@ def _duplicate_items(items: t.List[_T]) -> t.List[_T]:
...
+            if hasattr(item, 'callspec'):
+                target = item.callspec.params.get('target', None)
+            if target == 'auto' or target is None:
+                logging.warning(
+                    f"Test item '{item.originalname}': "
+                    "Ensure 'target' is included in the test @pytest.mark.parametrize when using 'skip_if_soc', "
+                    "or provide the --target argument when running tests (except 'auto' and multi-dut)."
+                )
+                continue
+            if '|' in target:
```
horw, line 1272: There is a question about multi-DUT support: should we support it? If yes, I think we need to do something like masking -> `esp32|esp32c5`, where the mask would be `0|1`, or for `esp32c5|esp32c6`, use the default mask(by default check all), or for `esp32c5|esp32`, mask it as`1|0`.

ex:
```python
@pytest.mark.skip_if_soc("SOC_BLE_SUPPORTED == 1", target_check_mask='1|0')

```

hfudev: nice catch. how about an api like `..., check_on_duts=[0, 1]` or `check_on_duts='all'` (default)?

igrr: I think such case is rare enough that we can simply move the "skip" check into the test case itself. And then we don't have to introduce or handle an additional argument which 99% of the time won't be used.

pytest-embedded-idf/tests/test_idf.py
```diff
@@ -987,3 +987,54 @@ def test_python_case(dut):
...
+
+    def run_pytest_with_target(target):
+        count = len(target.split('|'))
+        return testdir.runpytest(*EMBEDDED_SERVICES, '--target', target, '--count', count)
+
+    testdir.makepyfile("""
+        import pytest
+        from idf_build_apps.constants import SUPPORTED_TARGETS
+
+        @pytest.mark.skip_if_soc("SOC_ULP_LP_UART_SUPPORTED == 1")
```
igrr, line 1003: Are more complex conditions supported, or do we have the same limitation as in idf-build-apps, where we can't write `A == 1 or B == 2 or C == 3`?

hfudev: good point. I'll fix this limitation together while moving the if parser to a separate package.

pytest-embedded/pytest_embedded/plugin.py
horw: Ok, so let's not add this overwhelm here; just raise a warning if there few dut.

pytest-embedded-idf/pyproject.toml
```diff
@@ -33,6 +33,7 @@ requires-python = ">=3.7"
 dependencies = [
     "pytest-embedded~=1.12.1",
     "esp-idf-panic-decoder",
+    "idf-build-apps~=2.5.3"
```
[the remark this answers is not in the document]
horw: We created a separate package and moved this logic to `esp-bool-parser`.

pytest-embedded-idf/tests/test_idf.py
horw: This limitation has been removed in `esp-bool-parser`.

pytest-embedded-idf/pyproject.toml
```diff
@@ -33,6 +33,7 @@ requires-python = ">=3.7"
 dependencies = [
     "pytest-embedded~=1.12.1",
     "esp-idf-panic-decoder",
+    "esp-bool-parser~=0.1.0"
```
hfudev, line 36:
```suggestion
    "esp-bool-parser~=0.1"
```

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1252,7 +1253,35 @@ def _duplicate_items(items: t.List[_T]) -> t.List[_T]:
...
+                logging.warning("Multi-dut tests do not support the 'skip_if_soc' marker.")
+                continue
+
+            from esp_bool_parser import parse_bool_expr
```
hfudev, line 1276: why import here?

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1252,7 +1253,35 @@ def _duplicate_items(items: t.List[_T]) -> t.List[_T]:
...
+
+            stm = parse_bool_expr(skip_marker.args[0])
+            if not stm.get_value(target, ''):
+                from _pytest.mark import MARK_GEN
+
+                reason = f'Filtered by {skip_marker.args[0]}, for {target}.'
+                item.add_marker(MARK_GEN.skip(reason=reason))
```
hfudev, line 1283:
```suggestion
                reason = f'Filtered by {skip_marker.args[0]}, for {target}.'
                item.add_marker(pytest.mark.skip(reason=reason))
```
this is more like their "public api"?

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1252,7 +1253,35 @@ def _duplicate_items(items: t.List[_T]) -> t.List[_T]:
...
+        for item in items:
+            skip_marker = item.get_closest_marker('skip_if_soc')
+            if not skip_marker:
+                continue
+
+            target = config.getoption('--target', None)
+            if hasattr(item, 'callspec'):
+                target = item.callspec.params.get('target', None)
+            if target == 'auto' or not isinstance(target, str):
+                logging.warning(
```
hfudev, line 1266: we shall say something like `Ignoring pytest.mark.skip_if_soc, because ...`

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1252,7 +1253,35 @@ def _duplicate_items(items: t.List[_T]) -> t.List[_T]:
...
+                    f"Test item '{item.originalname}': "
+                    "Ensure 'target' is included in the test @pytest.mark.parametrize when using 'skip_if_soc', "
+                    "or provide the --target argument when running tests (except 'auto' and multi-dut)."
+                )
+                continue
+            if '|' in target:
+                logging.warning("Multi-dut tests do not support the 'skip_if_soc' marker.")
```
hfudev, line 1273: same here.

pytest-embedded/pytest_embedded/plugin.py
hfudev: besides, maybe here we shall use `warnings.warn`, since it's more like a warning to the developer, to let them fix it, instead of a log for helping debugging.

hfudev (commented):
LGTM.

pytest-embedded/pytest_embedded/plugin.py
horw: Actually I did it correctly, as it should only be imported if `skip_if_soc` is used. So, I think it's fine to have it here. What I can do is move it closer to the `if` condition.

pytest-embedded/pytest_embedded/plugin.py
horw: yea, much better

pytest-embedded/pytest_embedded/plugin.py
horw: updated

pytest-embedded/pytest_embedded/plugin.py
horw: Updated

pytest-embedded-idf/pyproject.toml
horw: Let's use `esp-bool-parser>=0.1.2,<1` as in `idf-build-apps` instead.

horw:
Let's mock `esp_bool_parser`'s `SUPPORTED_TARGET`, and `SOC_CAPS`, because we don't rely on ESP-IDF now; let's just mock them.

hfudev:
> Let's mock `esp_bool_parser`'s `SUPPORTED_TARGET`, and `SOC_CAPS`, because we don't dependent on ESP-IDF repo itself; let's just mock them.

sure. we could create a temp `IDF_PATH` folder, and create `IDF_PATH/tools/cmake/version.cmake` and `IDF_PATH/components/soc/<target>/include/soc/soc_caps.h` to mock it.

tests/esp-idf/components/soc/esp32/include/soc/soc_caps.h
hfudev, line 1: I don't think that we shall upload all files with all content. we could write some mock ones, which would be easier for testing. for exmaple

```
#define SOC_INTEGER_1  1
#define SOC_STRING_FOO "foo"
```

then we could know what the soc caps is while writing code. otherwise when we're writing `test_skip_if_soc`, I couldn't tell if the test case should be skipped or not.

```
      @pytest.mark.skip_if_soc("SOC_ADC_SAMPLE_FREQ_THRES_HIGH == 83333")  # <-- should it pass?
        @pytest.mark.parametrize('target', SUPPORTED_TARGETS, indirect=True)
        def test_adc():
            pass
```

horw: You are right, doing it this way is easier to understand. Got it, thank you!

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1252,7 +1254,39 @@ def _duplicate_items(items: t.List[_T]) -> t.List[_T]:
...
+            if '|' in target:
+                warnings.warn(
+                    'Ignoring pytest.mark.skip_if_soc, '
+                    "because multi-DUT tests do not support the 'skip_if_soc' marker. "
+                    'Please adjust the test setup accordingly.'
+                )
+                continue
+
+            stm = parse_bool_expr(skip_marker.args[0])
+            if stm.get_value(target, ''):
```
horw, line 1286: In the previous version, I did `not`, but I think this is wrong. If this is true, then we skip our test case because of `skip_if`.

tests/esp-idf/components/soc/esp32/include/soc/soc_caps.h
horw: Fixed

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1252,7 +1254,39 @@ def _duplicate_items(items: t.List[_T]) -> t.List[_T]:
...
     @pytest.hookimpl(hookwrapper=True, trylast=True)
-    def pytest_collection_modifyitems(self, items: t.List[Function]):
+    def pytest_collection_modifyitems(self, config: Config, items: t.List[Function]):
+        for item in items:
+            skip_marker = item.get_closest_marker('skip_if_soc')
+            if not skip_marker:
+                continue
+
+            from esp_bool_parser import parse_bool_expr
+
```
hfudev, line 1264: shall we check the embedded_services enabled here? something like

```
if 'idf' not in embedded_services and skip_marker:
    raise ValueError('... must use together')  # or a warning and ignore the marker?

from esp_bool_parser import parse_bool_expr
```

otherwise the users may face unexpected error messages

hfudev (commented):
LGTM. please also add a documentation page under `usages/markers.rst`

pytest-embedded/pytest_embedded/plugin.py
horw: updated

.github/workflows/test-python.yml
```diff
@@ -63,5 +63,5 @@ jobs:
       - uses: actions/upload-artifact@v4
         if: failure()
         with:
-          name: logs-${{ python-version }}
+          name: logs-${{ matrix.python-version }}
```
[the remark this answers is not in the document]
horw: :+1:

docs/usages/markers.rst
```diff
@@ -0,0 +1,67 @@
...
+.. code:: python
+
+   import pytest
+   from esp_bool_parser.constants import SUPPORTED_TARGETS
+
+
+   @pytest.mark.skip_if_soc("{condition}")
+   @pytest.mark.parametrize("target", SUPPORTED_TARGETS, indirect=True)
+   def test_template():
+       pass
```
hfudev, line 67:
```suggestion
Here are examples of how to use ``skip_if_soc`` with different conditions:

**Condition 1**: A boolean expression such as ``SOC_ULP_SUPPORTED != 1 and SOC_UART_NUM != 3``. This skips tests for chips that:

   - Do not support the ``low power mode`` feature (``SOC_ULP_SUPPORTED != 1``).
   - **And** have a UART number other than 3 (``SOC_UART_NUM != 3``).

.. code:: python

   @pytest.mark.skip_if_soc("SOC_ULP_SUPPORTED != 1 and SOC_UART_NUM != 3")
   @pytest.mark.parametrize("target", ["esp32", "esp32s2", "esp32c3"], indirect=True)
   def test_template_first_condition():
       pass

---

**Condition 2**: A boolean expression such as ``SOC_ULP_SUPPORTED != 1 or SOC_UART_NUM != 3``. This skips tests for chips that:

   - Either do not support the ``low power mode`` feature (``SOC_ULP_SUPPORTED != 1``).
   - **Or** have a UART number other than 3 (``SOC_UART_NUM != 3``).

.. code:: python

   @pytest.mark.skip_if_soc("SOC_ULP_SUPPORTED != 1 or SOC_UART_NUM != 3")
   @pytest.mark.parametrize("target", ["esp32", "esp32s2", "esp32c3"], indirect=True)
   def test_template_second_condition():
       pass
       
---

**Condition 3**: You can use a shortcut to apply this condition to all ESP-IDF supported targets (assuming ``IDF_PATH`` is set).

.. code:: python

   import pytest
   from esp_bool_parser.constants import SUPPORTED_TARGETS

   @pytest.mark.skip_if_soc("SOC_ULP_SUPPORTED != 1")
   @pytest.mark.parametrize("target", SUPPORTED_TARGETS, indirect=True)
   def test_template():
       pass
```

hfudev: when we're giving code examples, better to make them copy-paste-able

horw: Makes sense. The code example I provided should be considered as something else, not an example.

PR #331, opened 2025-01-13
feat: parametrize_wrapper

horw:
## Description

Add idf_parametrize decorator to simplify pytest parametrization by handling target expansion and runner marks via string representation.

- Automatically expanding special target values ('supported_targets' and 'preview_targets') into individual test cases;
- Simplifying pytest mark usage by allowing string-based mark definitions instead of direct pytest.mark usage;
- Unifying the parametrization pattern across esp-idf test cases;

## Testing

---

## Checklist

### DIFF
--- a/docs/usages/markers.rst
+++ b/docs/usages/markers.rst
@@ -61,3 +61,227 @@ Here are examples of how to use ``skip_if_soc`` with different conditions:
    @pytest.mark.parametrize("target", SUPPORTED_TARGETS, indirect=True)
    def test_template():
        pass
+
+*********************
+ ``idf_parametrize``
+*********************
+
+``idf_parametrize`` is a wrapper around ``pytest.mark.parametrize`` that simplifies and extends string-based parameterization for tests. By using ``idf_parametrize``, testing parameters becomes more flexible and easier to maintain.
+
+**Key Features:**
+
+-  **Target Expansion**: Automatically expands lists of supported targets, reducing redundancy in test definitions.
+-  **Markers**: use a marker as one of the parameters. If a marker is used, put it as the last parameter.
+
+Use Cases
+=========
+
+Target Extension
+----------------
+
+In scenarios where the supported targets are [esp32, esp32c3, esp32s3], ``idf_parametrize`` simplifies the process of creating parameterized tests by automatically expanding the target list.
+
+**Example:**
+
+.. code:: python
+
+   @idf_parametrize('target', [
+       ('supported_targets'),
+   ], indirect=True)
+   @idf_parametrize('config', [
+       'default',
+       'psram'
+   ])
+   def test_st(dut: Dut) -> None:
+       ...
+
+**Equivalent to:**
+
+.. code:: python
+
+   @pytest.mark.parametrize('target', [
+       'esp32',
+       'esp32c3',
+       'esp32s3'
+   ], indirect=True)
+   @pytest.mark.parametrize('config', [
+       'default',
+       'psram'
+   ])
+   def test_st(dut: Dut) -> None:
+       ...
+
+**Resulting Parameters Matrix:**
+
+.. list-table::
+   :header-rows: 1
+
+   -  -  Target
+      -  Config
+   -  -  esp32
+      -  default
+   -  -  esp32c3
+      -  default
+   -  -  esp32s3
+      -  default
+   -  -  esp32
+      -  psram
+   -  -  esp32c3
+      -  psram
+   -  -  esp32s3
+      -  psram
+
+Markers
+-------
+
+Markers can also be combined for added flexibility. It must be placed in the last position. In this case, if some test cases do not have markers, you can skip their definition. Look at the example.
+
+**Example:**
+
+In IDF testing, an environment marker (``marker``) determines which test runner will execute a test. This enables tests to run on various runners, such as:
+
+-  **generic**: Tests run on generic runners.
+-  **sdcard**: Tests require an SD card runner.
+-  **usb_device**: Tests require a USB device runner.
+
+.. code:: python
+
+   @pytest.mark.generic
+   @idf_parametrize('config', [
+       'defaults'
+   ], indirect=['config'])
+   @idf_parametrize('target, markers', [
+       ('esp32', (pytest.mark.usb_device,)),
+       ('esp32c3')
+       ('esp32', (pytest.mark.sdcard,))
+   ], indirect=['target'])
+   def test_console(dut: Dut, test_on: str) -> None:
+     ...
+
+**Resulting Parameters Matrix:**
+
+.. list-table::
+   :header-rows: 1
+
+   -  -  Target
+      -  Markers
+   -  -  esp32
+      -  generic, usb_device
+   -  -  esp32c3
+      -  generic, sdcard
+   -  -  esp32
+      -  generic, sdcard
+
+Examples
+========
+
+Target with Config
+------------------
+
+**Example:**
+
+.. code:: python
+
+   @idf_parametrize('target, config', [
+       ('esp32', 'release'),
+       ('esp32c3', 'default'),
+       ('supported_target', 'psram')
+   ], indirect=True)
+   def test_st(dut: Dut) -> None:
+       ...
+
+**Resulting Parameters Matrix:**
+
+.. list-table::
+   :header-rows: 1
+
+   -  -  Target
+      -  Config
+   -  -  esp32
+      -  release
+   -  -  esp32c3
+      -  default
+   -  -  esp32
+      -  psram
+   -  -  esp32c3
+      -  psram
+   -  -  esp32s3
+      -  psram
+
+Supported Target on Runners
+---------------------------
+
+**Example:**
+
+.. code:: python
+
+   @idf_parametrize('target, markers', [
+       ('esp32', (pytest.mark.generic, )),
+       ('esp32c3', (pytest.mark.sdcard, )),
+       ('supported_target', (pytest.mark.usb_device, ))
+   ], indirect=True)
+   def test_st(dut: Dut) -> None:
+       ...
+
+**Resulting Parameters Matrix:**
+
+.. list-table::
+   :header-rows: 1
+
+   -  -  Target
+      -  Markers
+   -  -  esp32
+      -  generic
+   -  -  esp32c3
+      -  sdcard
+   -  -  esp32
+      -  usb_device
+   -  -  esp32c3
+      -  usb_device
+   -  -  esp32s3
+      -  usb_device
+
+Runner for All Tests
+--------------------
+
+**Example:**
+
+.. code:: python
+
+   @pytest.mark.generic
+   @idf_parametrize('target, config', [
+       ('esp32', 'release'),
+       ('esp32c3', 'default'),
+       ('supported_target', 'psram')
+   ], indirect=True)
+   def test_st(dut: Dut) -> None:
+       ...
+
+**Resulting Parameters Matrix:**
+
+.. list-table::
+   :header-rows: 1
+
+   -  -  Target
+      -  Config
+      -  Markers
+
+   -  -  esp32
+      -  release
+      -  generic
+
+   -  -  esp32c3
+      -  default
+      -  generic
+
+   -  -  esp32
+      -  psram
+      -  generic
+
+   -  -  esp32c3
+      -  psram
+      -  generic
+
+   -  -  esp32s3
+      -  psram
+      -  generic
--- a/pytest-embedded-idf/pytest_embedded_idf/unity.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/unity.py
@@ -1,0 +1,80 @@
+import typing as t
+
+import pytest
+from esp_bool_parser import PREVIEW_TARGETS, SUPPORTED_TARGETS
+
+
+def _expand_target_values(values: t.List[t.List[t.Any]], target_index: int) -> t.List[t.List[t.Any]]:
+    """
+    Expands target-specific values into individual test cases.
+    """
+    expanded_values = []
+    for value in values:
+        target = value[target_index]
+        if target == 'supported_targets':
+            expanded_values.extend([
+                value[:target_index] + [target] + value[target_index + 1 :] for target in SUPPORTED_TARGETS
+            ])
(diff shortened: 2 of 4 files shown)


horw:
Let's add test cases for `_expand_target_values`, `_process_pytest_value`, and `idf_parametrize` itself. If you think some other cases should be covered, let me know, and I will add them.

pytest-embedded-idf/pytest_embedded_idf/unity.py
```diff
@@ -0,0 +1,80 @@
...
+            if isinstance(element, tuple):
+                marks.extend(getattr(pytest.mark, mark) for mark in element)
+        else:
+            params.append(element)
+
+    return pytest.param(*params, marks=marks)
+
+
+def idf_parametrize(
+    param_names: str, values: t.List[t.Union[t.Any, t.Tuple[t.Any, ...]]], indirect: bool = False
```
hfudev, line 48: I don't like the API design here, it's not intuitive for our users. Could we make it something like

```python
def idf_parametrize(
    *,
    targets: ...,
    temp_skip_ci_targets: ... = None,
    temp_skip_targets: ... = None,
    markers: ... = None
    indirect=True,
    **kwargs,
)
    ...
```

So our users know what to pass, and the arguments are more straightforward?

pytest-embedded-idf/pytest_embedded_idf/unity.py
```diff
@@ -0,0 +1,80 @@
...
+        indirect (bool): If True, marks parameters as indirect.
+
+    Returns:
+        Decorated test function with parametrization applied
+    """
+    param_list = [name.strip() for name in param_names.split(',')]
+    if not param_list:
+        raise ValueError('No parameter names provided')
+
+    runner_index = param_list.index('mark_runners') if 'mark_runners' in param_list else -1
```
hfudev, line 65: here you mean `env_marker`?

hfudev (commented):
Thanks for the PR, please also update our user guide with some examples!

pytest-embedded-idf/pytest_embedded_idf/unity.py
horw: Yes, this is ENV_MARKERS

pytest-embedded-idf/pytest_embedded_idf/unity.py
horw: What about arrays? I don’t think adding new values will be easy.

hfudev: we could make all `kwargs` as param. for exmaple,

```python
@idf_parametrize(targets=['esp32', 'esp32s2'], custom_var=[['1', '2'], ['3', 4']])
```

will turn into

```python
@pytest.mark.idf_parametrize('targets, custom_var', [
    ('esp32', ['1', '2']),
    ('esp32', ['1', '2']),
    ('esp32', ['1', '2']),
    ('esp32', ['1', '2']),

Comment (horw) [reply pytest-embedded-idf/pytest_embedded_idf/unity.py:48]: Hmm, okay, let's do it like this then.

Comment (horw) [reply pytest-embedded-idf/pytest_embedded_idf/unity.py:48]: But we can't use a multiplier here because not all values can be multiplied...

Comment (horw) [reply pytest-embedded-idf/pytest_embedded_idf/unity.py:48]: It’s quite tricky to find a good solution that fits all our requirements. However, it should at least be capable of doing what `pytest.mark.parametrize` is able to do.

For example, the current solution can do the following:
```python
@idf_parametrize("target, a, b", [
    ("supported", 2, 3),
])
```
which will be transformed into:
```python
@pytest.mark.parametrize("target, a, b", [
    ("esp32", 2, 3),
    ("esp32c3", 2, 3),
    ("esp32s3", 2, 3),
    ("...", 2, 3),
])
```

Additionally, it can use a `pytest.mark.parametrize` multiplier to achieve something like this:
```python
@pytest.mark.parametrize("a, b", [
    (2, 3),
])
@pytest.mark.parametrize("target", [
    "esp32",
    "esp32c3",
    "...",
])
def test()
```
which will be equivalent to:
```python
@idf_parametrize("a, b", [
    (2, 3),
])
@idf_parametrize("target", [
    "supported",
])
def test()
```

However, the current solution is not very user-friendly, as `pytest.mark.parametrize`, in general, can also be a bit tricky to use 😅.

We need to consider something better. @hfudev provided a very good idea to use `args` and `kwargs`, but this approach would involve considering how to handle multiplication within the function. This would require a multiplier inside the function, which could complicate separating what needs to be multiplied from what doesn’t.

For example, consider this case:
```python
def idf_parametrize(
    targets=["esp32", "esp32c3"],
    a=[1, 2],
    b=[3],
):
    ...
```
The result would be:
- `(esp32, 1, 3)`
- `(esp32c3, 1, 3)`
- `(esp32, 2, 3)`
- `(esp32c3, 2, 3)`

But for some reason, I don’t need `(esp32c3, 2, 3)`. How can I avoid it? It becomes tricky.  

If we use this as part of the previous solution but reorganize the order of parameters (not like in `pytest.mark.parametrize` but with `args`), it would look like this:
```python
def idf_parametrize(
    targets=["esp32", "esp32c3", "esp32p4"],
    a=[1, 2, 1],
    b=[3, 4, 2],
):
    ...
```
The result would be:
- `(esp32, 1, 3)`
- `(esp32c3, 2, 4)`
- `(esp32p4, 1, 2)`

But there’s also a downside:
This approach becomes tricky when adding new parameters because they need to match the array size. If you add just one value, you’d have to add it __vertically__ to all args. In comparison, `pytest.mark.parametrize`'s __horizontal__ approach still seems better.

I believe the main point lies in multiplication: making this function handle multiplication effectively. However, neither my solution nor the current one achieves the ideal design.

hfudev: I came up with another idea as follows, but it makes things even more complicated...

```
@idf_parametrize(
    targets=[
        "esp32",
        pytest.param("esp32c3", markers=[...]),  # support pytest.param here
    ],
    params={
        "a": [1, 2],
        "b": [3, 4],
    },
    exclude=[
        {"target": "esp32c3", "a": 1, "b": 4}
    ],
    include=[
        {"target": "esp32", "a": 5, "b": 6}
    ]
)
def test_function(target, a, b):
    print(f"Running test with target={target}, a={a}, b={b}")
```

and will expand to

| target | a | b |
| --- | --- | --- |
| esp32 | 1 | 3 |
| esp32 | 1 | 4 |
| esp32 | 2 | 3 |
| esp32 | 2 | 4 |
| included by `include` esp32 | 5 | 6 |
| esp32c3 | 1 | 3 |
| excluded by `exclude` ~~esp32c3~~ | ~~1~~ | ~~4~~ |
| esp32c3 | 2 | 3 |
| esp32c3 | 2 | 4 |

hfudev: let's follow your original design, at least it's almost the same behavior as the original `pytest.mark.parametrize` so users dont have to learn it twice.

Please describe all the possible usages under `docs/en/usages/markers.rst`.

docs/usages/markers.rst
```diff
@@ -0,0 +1,291 @@
...
+
+*********************
+ ``idf_parametrize``
+*********************
+
+``idf_parametrize`` is a wrapper around ``pytest.mark.parametrize`` that simplifies and extends string-based parameterization for tests. By using ``idf_parametrize``, testing parameters becomes more flexible and easier to maintain.
+
+**Key Features:**
+
+-  **Target Expansion**: Automatically expands lists of supported targets, reducing redundancy in test definitions.
```
horw, line 73: Should we add some extra functionality?
I mean, should we be able to parse other marks or something else?

docs/usages/markers.rst
```diff
@@ -61,3 +61,231 @@ Here are examples of how to use ``skip_if_soc`` with different conditions:
...
+
+   @idf_parametrize('env_marker', [
+       ('generic'),
+       ('sdcard', 'usb_device'),
+       'sdcard',
+       'sdcard, generic'
+   ], indirect=True)
+   @idf_parametrize('target', [
+       'esp32'
+   ])
```
hfudev, line 157: let's try support real markers. like

```suggestion
   @idf_parametrize('env_marker', [
       (pytest.mark.skip),
       (pytest.mark.foo, pytest.mark.bar),
   ], indirect=True)
   @idf_parametrize('target', [
       'esp32'
   ])
```

not sure if it's feasible.

horw: updated

horw:
@hfudev I have updated and added tests for this PR; please take a look.

hfudev:
The PR LGTM. A few notes here:
- no need to touch the unmodified files, those format-related changes could be done when the test scripts are updated. (but let's leave them as they are, don't waste your time reverting them)
- please add the `idf_parametrize` in `pytest_embedded_idf.__all__`.

and let's hold this PR for a while, until the internal formatting MR has been updated by this. some potential bugs or features may need to be added.

horw:
@hfudev I think about this one:  

```
@pytest.mark.generic
@idf_parametrize('config,markers,target', [
    ('default', (), 'supported_targets'),
    ('freertos_options', (), 'supported_targets'),
    ('psram', (), 'esp32'),
    ('psram', (), 'esp32c5'),
    ('psram', (), 'esp32p4'),
    ('psram', (), 'esp32s3'),
    ('single_core', (), 'esp32'),
    ('single_core', (), 'esp32p4'),
    ('smp', (pytest.mark.temp_skip_ci(targets=['esp32p4', 'esp32c5', 'esp32c61'], reason='test failed/TBD IDF-8113'),), 'supported_targets')
], indirect=True)
```

Now markers look like this.  
I think we can do it better.  
Just remove 'markers' from 'config,markers,target' and directly try to detect an array with marks.

```
@pytest.mark.generic
@idf_parametrize('config,target', [
    ('default', 'supported_targets'),
    ('freertos_options', 'supported_targets'),
    ('psram', 'esp32'),
    ('psram', 'esp32c5'),
    ('psram', 'esp32p4'),
    ('psram', 'esp32s3'),
    ('single_core', 'esp32'),
    ('single_core', 'esp32p4'),
    ('smp', (pytest.mark.temp_skip_ci(targets=['esp32p4', 'esp32c5', 'esp32c61'], reason='test failed/TBD IDF-8113'),), 'supported_targets'),
    ((pytest.mark.temp_skip_ci(targets=['esp32p4', 'esp32c5', 'esp32c61'], reason='test failed/TBD IDF-8113'),), 'smp', 'supported_targets')
], indirect=True)
```

hfudev:
how about we force the markers must be located as the last argument? to avoid confusion.

docs/usages/markers.rst
```diff
@@ -61,3 +61,231 @@ Here are examples of how to use ``skip_if_soc`` with different conditions:
...
+      -  psram
+
+Supported Target on Runners
+---------------------------
+
+**Example:**
+
+.. code:: python
+
+   @idf_parametrize('target, markers', [
```
hfudev, line 220: according to the code, here should be ?

```suggestion
   @idf_parametrize('target, __markers', [
```

---

btw, what's the conclusion of the earlier discussion.

can markers be the last one implicitly? without declare it in the param list?

hfudev (commented):
Overall LGTM.

but I'm a bit confused... shall I use `markers` or `__markers`...

your tests and docs are written as `markers`, but in `utils.py` it's written like `__markers`.

besides, can we set it implicitly? Just asking for confirmation of the earlier question, I actually prefer setting it explicitly with spelling `markers`.

docs/usages/markers.rst
horw: Because I allowed a single parameter to be used without an array, for example:  

```python
@idf_parametrize('config', [
   'a',
   'b',
   'c'
])
```

When I want to mix a single parameter with a marker, there is an issue—it should be an array, like this:  

```python
@idf_parametrize('config', [ # --> should be 'config, marker'
   ('a', (marker,)),
   ('b', (marker,)),
   ('c', (marker,)),
])
```

When there is more than one parameter, it is straightforward. However, with only one parameter, the `marker` must be explicitly included. If there is more than one parameter, the marker can be placed in the last positional argument, without put `marker` in parameters.

hfudev: so the correct spell here is `markers`?

> the marker can be placed in the last positional argument, without put marker in parameters.

let's make `markers` always have to be declared explicitly. it's important to keep the behaviors consistent.

pytest-embedded-idf/pytest_embedded_idf/utils.py
```diff
@@ -0,0 +1,88 @@
...
+    Returns:
+        Decorated test function with parametrization applied
+    """
+    param_list = [name.strip() for name in param_names.split(',')]
+    for param in param_list:
+        if not param:
+            raise ValueError(f'One of the provided parameters name is empty: {param_list}')
+
+    param_count = len(param_list)
+    param_list[:] = [_p for _p in param_list if _p not in ('__markers',)]
```
hfudev, line 75: here.

docs/usages/markers.rst
horw: Okay, let's name it markers and always place it in the last position.

hfudev: if we set it explicitly, why we need to limit the position?

pytest-embedded-idf/pytest_embedded_idf/utils.py
horw: updated

docs/usages/markers.rst
horw:
```
@idf_parametrize('config,marker', [
   ('a', (marker,)),
   ('b',),
   ('c',),
])
```

hfudev: nice. makes sense. could you add this use case to the documentation as well?

horw: Updated docs, thank you for noticing these issues.

hfudev:
@horw PR LGTM. however your test cases are failing....

https://github.com/espressif/pytest-embedded/actions/runs/13255583811/job/37001800796?pr=331#step:6:14315

hfudev:
LGTM! I'll make a release soon

PR #333, opened 2025-01-17
feat: support --add-target-as-marker-with-amount

hfudev:


pytest-embedded/tests/test_base.py
```diff
@@ -626,9 +641,82 @@ def test_unclosed_file_handler(test_input, dut):
...
+                pass
+        """)
+
+        result = pytester.runpytest('--add-target-as-marker-with-amount', 'y', '-vvvv')
+
+        result.assert_outcomes(passed=3)
+        result.stdout.fnmatch_lines([
+            '*Unknown pytest.mark.esp32+esp8266 - is this a typo?*',
+            '*Unknown pytest.mark.esp32_2 - is this a typo?*',
+            '*Unknown pytest.mark.esp32+esp32s2+esp8266 - is this a typo?*',
```
horw, line 710: So, we still want to extend our functionality with markers?

hfudev: this feature is used for selecting the runners while running the tests in CI. the markers will be added when `target` is defined in `param`.

Issue #332, opened 2025-01-17
Record `app_path` in Junit Reports (RDT-1098)


hfudev:
terms:

- main junit report -> the junit file generated by `--junitxml`
- dut junit reports -> the junit files generated for multi-dut tests `dut-0.xml`, `dut-1.xml`, ...

For single-dut test cases, we shall record the `app_path` into the main junit report

For multi-dut test cases, we shall first record them correctly into their own dut junit reports, then use the `JunitMerger` to copy those to the main junit report. For example, 

- dut[0].app_path -> app_path[0] -> dut-0.xml
- dut[1].app_path -> app_path[1] -> dut-1.xml

for `app_path`, it's better to use relative path to the pytest rootdir.

hfudev:
closed by #338

PR #336, opened 2025-01-22
feat(junit): record app_path in JUnit reports

alekseiapa:
## Description

This PR introduces the feature to record attribute `app_path` in JUnit reports.

## Related

Closes https://github.com/espressif/pytest-embedded/issues/332

## Testing

Added test that ensures that the `app_path` attribute is correctly recorded for each `<testcase>` element in the JUnit report.

---

## Checklist

- [x] All CI checks (GH Actions) pass.


alekseiapa:
Hi @hfudev . Please take a look at the changes. Thanks!

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1273,6 +1298,25 @@ def get_param(item: Function, key: str, default: t.Any = None) -> t.Any:
...
+            tree = ET.parse(file)
+            root = tree.getroot()
+
+            for testcase in root.findall('.//testcase'):
+                testcase_name = testcase.get('name')
+                if testcase_name in testcase_app_paths_map:
+                    relative_paths = {os.path.relpath(path, rootdir) for path in testcase_app_paths_map[testcase_name]}
+                    testcase.set('app_path', ':'.join(relative_paths))
+
+            buffer = StringIO()
```
horw, line 1313: What do you think about reusing this instead of the buffer?
```suggestion
      fw.write(escape_illegal_xml_chars(ET.tostring(root, encoding='unicode')))
```

alekseiapa: I believe using` ElementTree.write` with a buffer is a safer  approach. After` ET.tostring()` creates the XML string, `escape_illegal_xml_chars()` is applied to clean it up by removing illegal characters. However, if the input to `ET.tostring()` already contains illegal characters, they will appear in the serialized string and need to be removed.

horw:
Shall we also add a test for `test_idf.py` to ensure it works with apps?

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1167,12 +1168,34 @@ def dut(
...
+        serial=serial,
+        qemu=qemu,
+        wokwi=wokwi,
+    )
+
+    if _testcase_app_paths_map_key not in request.config.stash:
+        request.config.stash[_testcase_app_paths_map_key] = {}
+
+    stash = request.config.stash[_testcase_app_paths_map_key]
+    testsuite_name = dut_instance.test_case_name
```
hfudev, line 1191: nitpick. testsuite means something different in junit report field.

usually something like 

```
    <testsuite name="pytest" errors="0" failures="1" skipped="0" tests="1" time="33.576" timestamp="2025-01-23T14:07:31.637710+01:00">
```

the parent concept of all test cases.

you may just call `dut_instance.test_case_name` directly

pytest-embedded/pytest_embedded/unity.py
```diff
@@ -324,5 +326,11 @@ def merge(self, junit_files: List[str]):
             if int(junit_parent.attrib['failures']) > 0:
                 self.failed = True
 
+        for testcase in self.junit.findall('.//testcase'):
+            # Add the 'pytest_case_name' attribute for each test case to enable later extraction
+            # of the app_path from the pytest_case_name to app_path mapping
+            if 'pytest_case_name' not in testcase.attrib:
+                testcase.attrib['pytest_case_name'] = testcase.attrib.get('name')
```
hfudev, line 333: question, why can't we use the `attrib['name']` here?

pytest-embedded/pytest_embedded/plugin.py
```diff
@@ -1273,6 +1297,29 @@ def get_param(item: Function, key: str, default: t.Any = None) -> t.Any:
...
+            root = tree.getroot()
+
+            for testcase in root.findall('.//testcase'):
+                pytest_testcase_name = testcase.get('pytest_case_name')
+                if pytest_testcase_name in testcase_app_paths_map:
+                    # Map the app paths as relative paths and add to the 'app_path' attribute
+                    relative_paths = {
+                        os.path.relpath(path, rootdir) for path in testcase_app_paths_map[pytest_testcase_name]
+                    }
+                    testcase.set('app_path', ':'.join(relative_paths))
```
hfudev, line 1313:
```suggestion
                    testcase.set('app_path', ';'.join(relative_paths))
```

better use `;` here. since it's more seldomly used in paths... windows have `:` as drive separator.

hfudev (commented):
overall LGTM. a few comments posted

pytest-embedded/pytest_embedded/unity.py
alekseiapa: The issue is that in some cases, the pytest test case name is being replaced by the C test case name, resulting in the name being absent in the mapping.

alekseiapa: Unfortunately, this suggestion isn't viable. For example,` self.testsuite.add_unity_test_cases` is executed only within the `expect_unity_test_output` context. However, there are numerous scenarios where users call `expect_exact`. Additionally, I tested it on the provided tests in this PR, and in both cases, the proposed solution either fails to work or requires significantly more changes compared to the current implementation.

hfudev: I made this work in #338, please take a look. I'll close this PR.

PR #337, opened 2025-01-22
fix: read all chunks in queue since we have interval in listener

hfudev:
related to #334

### DIFF
--- a/pytest-embedded/pytest_embedded/dut_factory.py
+++ b/pytest-embedded/pytest_embedded/dut_factory.py
@@ -52,15 +52,16 @@ def msg_queue_gn() -> MessageQueue:
 def _listen(q: MessageQueue, filepath: str, with_timestamp: bool = True, count: int = 1, total: int = 1) -> None:
     _added_prefix = False
     while True:
-        msg = q.get()
-        if not msg:
+        msgs = q.get_all()
+        if not msgs:
             continue
 
+        msg_b = b''.join(msgs)
         with open(filepath, 'ab') as fw:
-            fw.write(msg)
+            fw.write(msg_b)
             fw.flush()
 
-        _s = to_str(msg)
+        _s = to_str(msg_b)
         if not _s:
             continue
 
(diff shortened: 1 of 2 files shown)


pytest-embedded/pytest_embedded/log.py
```diff
@@ -42,6 +44,10 @@ def put(self, obj, **kwargs):
         except:  # noqa # queue might be closed
             pass
 
+    def get_all(self) -> List[bytes]:
+        with self.lock:
```
horw, line 48: Could there be another process sharing this queue? If not, is it important to add a lock here?

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -52,15 +52,16 @@ def msg_queue_gn() -> MessageQueue:
 def _listen(q: MessageQueue, filepath: str, with_timestamp: bool = True, count: int = 1, total: int = 1) -> None:
     _added_prefix = False
     while True:
-        msg = q.get()
-        if not msg:
+        msgs = q.get_all()
+        if not msgs:
```
horw, line 56: This line will never be True.

pytest-embedded/pytest_embedded/log.py
```diff
@@ -42,6 +44,10 @@ def put(self, obj, **kwargs):
...
+            return [self.get() for _ in range(self.qsize())]
```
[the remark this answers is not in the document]
horw: https://github.com/python/cpython/issues/101094

pytest-embedded/pytest_embedded/log.py
hfudev: the queue is shared in different processes, so it's better to leave it there. In case other process called the `get_all` one day.

pytest-embedded/pytest_embedded/log.py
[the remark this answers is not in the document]
hfudev: nice point. i'll use the try except `get` then

erhankur:
@hfudev Thank you for looking into this. My original issue still exists, but at least the 100ms sleep effect will be resolved.

PR #344, opened 2025-02-27
fix(pytest-embedded-qemu): Allowed QEMU RISCV flash image sizes (RDT-1177)

Harshal5:
## Description

Running a pytest test case using QEMU for ESP32-C3, produces the following error:
```
2025-02-27 20:22:56 INFO Executing qemu-system-riscv32 -nographic -machine esp32c3 -qmp tcp:127.0.0.1:53445,server,wait=off -drive file=/Users/harshal/esp/esp-idf/tools/test_apps/security/secure_boot/test/esp32c3_efuses.bin,if=none,format=raw,id=efuse -global driver=nvram.esp32c3.efuse,property=drive,value=efuse -global driver=timer.esp32c3.timg,property=wdt_disable,value=true -drive file=/Users/harshal/esp/esp-idf/tools/test_apps/security/secure_boot/build/flash_image.bin,if=mtd,format=raw
2025-02-27 20:22:56 esptool.py v4.9.dev5
2025-02-27 20:22:56 SHA digest in image updated
2025-02-27 20:22:56 Wrote 0x80000 bytes to file /Users/harshal/esp/esp-idf/tools/test_apps/security/secure_boot/build/flash_image.bin, ready to flash to offset 0x0
2025-02-27 20:22:56 
2025-02-27 20:22:56 Adding SPI flash device
2025-02-27 20:22:56 qemu-system-riscv32: Drive size error: only 2, 4, 8, and 16MB images are supported
2025-02-27 20:22:56 qemu-system-riscv32: -drive file=/Users/harshal/esp/esp-idf/tools/test_apps/security/secure_boot/build/flash_image.bin,if=mtd,format=raw: machine type does not support if=mtd,bus=0,unit=0
```

ESP32-C3 (riscv) QEMU seems to support only 2, 4, 8, and 16MB flash image sizes [reference](https://github.com/espressif/qemu/blob/esp-develop/hw/riscv/esp32c3.c#L204-L220).

## Testing

---

## Checklist

### DIFF
--- a/pytest-embedded-qemu/pytest_embedded_qemu/app.py
+++ b/pytest-embedded-qemu/pytest_embedded_qemu/app.py
@@ -25,16 +25,10 @@ class IdfFlashImageMaker:
     ]
 
     RISCV_FLASH_BIN_SIZES: t.ClassVar[t.List[t.Tuple[int, str]]] = [
-        (256 * 1024, '256KB'),
-        (512 * 1024, '512KB'),
-        (1 * 1024 * 1024, '1MB'),
         (2 * 1024 * 1024, '2MB'),
         (4 * 1024 * 1024, '4MB'),
         (8 * 1024 * 1024, '8MB'),
         (16 * 1024 * 1024, '16MB'),
-        (32 * 1024 * 1024, '32MB'),
-        (64 * 1024 * 1024, '64MB'),
-        (128 * 1024 * 1024, '128MB'),
     ]
 
     def __init__(self, app: 'QemuApp', image_path: str, *, qemu_version: Version = Version('8.0.0')):


hfudev:
LGTM. Thank you for the fix!

igrr:
At this point, does it make sense to keep the supported flash sizes list different between Xtensa and RISC-V?

In QEMU, support of flash size for the specific machine is anyways implemented [in each machine's source file](https://github.com/espressif/qemu/blob/7591824ec477d9ae35d679bbc726d6f9f4ecffd2/hw/xtensa/esp32s3.c#L247-L270), not once per architecture. So even for the same architecture, the lists could be different... So if we are making the lists different, then one list per target would make more sense than one per architecture.

Harshal5:
> So if we are making the lists different, then one list per target would make more sense than one per architecture.

True!

hfudev:
@igrr I'll check how to improve this. Thanks for your insights!

PR #345, opened 2025-03-12
feat: get targets filtered by soc attrs

horw:
## Description

In this PR, a function is added that will return a target list according to the **SOC_CAPS** attribute. We already have `skip_if_soc`, but in some cases, it is better to directly filter by target rather than adding a skip mark for it.

## Testing

The `test_soc_filtered_targets` was added.

## Checklist

### DIFF
--- a/pytest-embedded-idf/pytest_embedded_idf/utils.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/utils.py
@@ -3,6 +3,7 @@ from contextvars import ContextVar
 
 import pytest
 from esp_bool_parser import PREVIEW_TARGETS, SUPPORTED_TARGETS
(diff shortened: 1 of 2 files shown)


horw:
@hfudev Hello! What do you think about this function? Should we add it to our utils or not, since we already have the supported_targets and preview_targets logic, getting the filtered target is quite easy.

hfudev:
I think it's a nice feature to have! Thank you for your idea.

could you check why the test cases are failing? I somehow assume that those are runner issues but could you double check.

if all looks good I'll merge it

horw:
> could you check why the test cases are failing? I somehow assume that those are runner issues but could you double check.


You are right; the failed test case is related to the target running the test. Not for this PR.

PR #378, opened 2025-10-08
Fix/hanging single dut cases breaks followup test cases

hfudev:


igrr:
@hfudev is it possible to add a test case to pytest-embedded/tests/fixtures/unit_test_app_esp32/main/case_tester_example.c (for example, entering an infinite loop) and check that it's handled correctly?

hfudev:
> @hfudev is it possible to add a test case to pytest-embedded/tests/fixtures/unit_test_app_esp32/main/case_tester_example.c (for example, entering an infinite loop) and check that it's handled correctly?

@igrr I was thinking about this too. Just not sure where to add it since the reset between each sub case will make it non-blocking. Will add test cases for sure.

PR #413, opened 2026-04-01
fix: support testing without flasher_args.json (CII-159)

Honza0297:
## Description

This allows testing build where only the ELF file is generated without flashable binaries. This can be the case e.g. for linux builds.

## Testing

---

## Checklist

### DIFF
--- a/pytest-embedded-idf/pytest_embedded_idf/dut.py
+++ b/pytest-embedded-idf/pytest_embedded_idf/dut.py
@@ -270,6 +270,11 @@ class IdfDut(IdfUnityDutMixin, SerialDut):
             # >>> self.gdb.write('load')
             return
 
+        if self.app.is_linux_elf:
+            # linux ELF file. OpenOCD is meaningless
+            logging.debug('Linux ELF-only build; skipping OpenOCD program_esp.')
+            return
+
         for _f in self.app.flash_files:
             if _f.encrypted:
                 raise ValueError("Encrypted files can't be flashed in via JTAG")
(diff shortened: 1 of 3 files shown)


pytest-embedded-idf/pytest_embedded_idf/serial.py
```diff
@@ -72,6 +72,10 @@ def _start(self):
         else:
             if self.app.is_loadable_elf:
                 self.load_ram()
+            elif getattr(self.app, 'is_linux_elf', False):
```
hfudev:
```suggestion
            elif self.app.is_linux_elf
```

pytest-embedded-idf/pytest_embedded_idf/dut.py
```diff
@@ -270,6 +270,11 @@ def flash_via_jtag(self):
             # >>> self.gdb.write('load')
             return
 
+        if getattr(self.app, 'is_linux_elf', False):
```
hfudev:
```suggestion
        if self.app.is_linux_elf:
```

hfudev (commented):
LGTM. These two places are all using `IdfApp`, which makes me think calling `self.app.is_linux_elf` directly is safe. WDYT?

Honza0297:
@hfudev Thanks for the review! You were right, I changed `getattr(self.app, 'is_linux_elf', False)` to `self.app.is_linux_elf`.

Honza0297:
By the way, any idea why the CI is failing? To be honest, I am not sure I understand why it fails in the first place... Thanks!

hfudev:
> By the way, any idea why the CI is failing? To be honest, I am not sure I understand why it fails in the first place... Thanks!

me neither...

but i think it's not related to your PR.

Thank you for the PR!

PR #416, opened 2026-04-17
fix(serial): during auto-reconnect, update all pySerial references (CII-181)

ofacklam:
## Description

The previous implementation of the auto-reconnection feature (commit 019e38a97d9d0be88569c50bd1426191a6f69010) detects connection loss in the reader _event_loop() and proceeds to re-assign the local variable to a fresh pySerial instance. However, the `proc` member of the Serial class is not updated, and attempts to write to the Serial instance continue to fail even after the reconnection logic succeeded.

Therefore, make _SerialRedirectThread take the Serial instance as argument, instead of the pySerial.Serial object.
The _event_loop() can thus operate on and update the `proc` member directly.

## Testing

Tested locally with an Espressif DevKit, manually disconnecting & reconnecting the serial line

---

## Checklist

### DIFF
--- a/pytest-embedded-serial/pytest_embedded_serial/serial.py
+++ b/pytest-embedded-serial/pytest_embedded_serial/serial.py
@@ -108,7 +108,7 @@ class Serial:
 
         # Here the reason why we're still using thread is,
         # the `pyserial` object can't be pickled when using multiprocessing.Process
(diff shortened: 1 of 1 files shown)


hfudev:
Thank you. LGTM!

@horw PTAL.

PR #423, opened 2026-06-12
feat: added env for setup test case default timeout

horw:
## Description

Actually, 90 seconds is quite a long timeout per test case, and when a test case starts failing, it's usually due to a timeout. So in most cases, the timeout should be around 10–20 seconds, and for some of them, we can explicitly set a higher timeout.


hfudev:
@horw could you make it as a cli option as well?

also please document the env var.

hfudev:
also the naming, probably should be `PYTEST_EMBEDDED_UNITY_DEFUALT_TIMEOUT`.

with some documentation

PR #426, opened 2026-08-06
fix: read the redirect log as bytes, not text (CII-240)

### LINKED ISSUE #425: `_forward_io` dies silently when a multi-byte UTF-8 char is split across writes, truncating DUT output (CII-239)
**Describe the bug**

`_PopenRedirectProcess._forward_io` (`pytest-embedded/pytest_embedded/log.py:155-162`, identical on v2.8.1 and `main`) reads the redirect log in text mode:

```python
with open(logfile) as fr:          # text mode, errors='strict'
    while True:
        try:
            msg_queue.put(fr.read())  # msg_queue may be closed
        except Exception:
            break
```

The file is being appended to concurrently, so the reader keeps hitting a temporary EOF. When that end falls inside a multi-byte sequence, the decoder raises:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2 in position 43: unexpected end of data
```

`except Exception` swallows it: the reader exits with `exitcode=0`, empty stderr, and DUT output simply stops mid-stream. Downstream this surfaces as `pexpect.exceptions.TIMEOUT` on whatever `expect()` was waiting for, with nothing anywhere mentioning decoding.

Not specific to `read()` without a size — `read(n)` and `readline()` fail the same way; only binary mode avoids it. The text-mode open exists in 1.x too, so the decode issue predates 387b877 (#374); that commit only made it silent (1.18.x prints a traceback and exits 1).

**Configuration Files**

Not relevant — no project config needed, and no ESP-IDF: the defect is in the pure-Python forwarding path.

**To Reproduce**

The trigger is a multi-byte character split across two writes to the log, which a byte-at-a-time UART/QEMU stream does naturally. Writing whole lines with a single `write()` does *not* reproduce.

The decode itself:

```python
p = 'x.log'
open(p, 'wb').write(b'hello \xe2\x80\x94 world\n[SKIP] tail \xe2')
open(p).read()   # UnicodeDecodeError: ... unexpected end of data
```

Against the real class: start `_PopenRedirectProcess(msg_queue, logfile)`, append an ASCII chunk (arrives), then append a chunk ending mid-em-dash — `is_alive=False`, `exitcode=0`, child stderr empty, and the text before the split byte is lost too (`read()` raises before returning anything). End to end in my run, one em-dash (U+2014) in device output stopped the stream at line 40 and the run failed with `TIMEOUT: Not found "Press ENTER to see the list of tests"` after 30 s.

**Expected behavior**

Forwarding should not depend on where chunk boundaries fall.

Suggested fix:

```python
with open(logfile, 'rb') as fr:
```

The decode looks unnecessary here: `MessageQueue.put` accepts `str | bytes` and calls `to_bytes()` (identity for bytes), `PexpectProcess`'s buffer is bytes, and the serial transport already puts raw bytes on the same queue (`pytest-embedded-serial/pytest_embedded_serial/serial.py:244`, `self._q.put(s)` with `s = self._s.read_all()`, no decode) — so today's path is a bytes → str → bytes round trip, and `'rb'` makes the two transports consistent.

One honest caveat: text mode does universal-newline translation, so `'rb'` stops swallowing CR (CRLF sample: 45 bytes written → 43 received before, 45 after). That is the same raw `\r\n` the serial path already delivers, and `_listen`'s `\r\n` → `\n` plus `splitlines()` in the Unity parser already handle it.

Optionally, `except Exception` could be narrowed — `MessageQueue.put` already swallows a closed queue itself, and the only escape I observed was `BrokenPipeError` at teardown — but I haven't proven that list exhaustive, so treat it as a suggestion.

Related, not part of this report: `pytest-embedded-idf/pytest_embedded_idf/unity_tester.py:181` does a strict `.decode('utf8')`, and `_listen`'s `to_str(msg)` uses `errors='ignore'`, so a split character vanishes from the console copy without a trace.

**Dev Environment (please complete the following information):**

- OS: Linux (Ubuntu 24.04, Docker) and macOS
- Python Version: 3.12.3 and 3.14.5
- Pytest Version: 9.1.1
- pytest-embedded Version: 2.8.1 and 2.2.1 (also 1.18.1, where it exits 1 with a traceback); `git diff v2.8.1..main -- pytest-embedded/pytest_embedded/log.py` is empty

I couldn't find an existing issue for this. Happy to send a PR with the `'rb'` change and a regression test if that's welcome.

hacker-cb:
Fixes #425.

`_PopenRedirectProcess._forward_io` read the redirect log in text mode while it was still
being appended to. Every read reaches a temporary EOF, which finalizes the incremental
decoder, so a UTF-8 sequence straddling that boundary raises `UnicodeDecodeError`. The
surrounding `except Exception` swallows it — the process exits 0 with an empty stderr and
DUT output stops for the rest of the session, surfacing later as a pexpect timeout that
names no cause.

## Why binary rather than a decoder

The decode was never needed here:

- `MessageQueue.put` accepts `str | bytes` and re-encodes through `to_bytes()`, which is
  the identity for bytes;
- `PexpectProcess`'s buffer is bytes;
- the serial transport already forwards `read_all()` undecoded
  (`pytest-embedded-serial/pytest_embedded_serial/serial.py:244`).

So today's path is a bytes → str → bytes round trip. Reading in binary removes the failure
mode and makes the two transports consistent.

## One behaviour change

Text mode applied universal-newline translation, so a DUT sending CRLF was forwarded as LF.
Binary mode preserves CR — which is what the serial transport already delivers, and both
`_listen` and the Unity parser handle CRLF already. `test_forward_io_preserves_carriage_returns`
pins it so it stays deliberate.

## Tests

Two unit tests next to the existing `test_listen_no_data_loss_*`, in the same shape.

The reproduction condition is worth noting: the multi-byte sequence has to be split
**between two appends**, which a byte-at-a-time UART/QEMU stream does naturally. Writing a
whole line in a single `write()` does not reproduce it.

On unmodified `main` both fail:

```
AssertionError: forwarder died on a split sequence (exitcode=0)
AssertionError: assert b'first\nsecond\n' == b'first\r\nsecond\r\n'
```

With this change both pass.

## Test runs on this machine

- `pytest-embedded/tests/` — 36 passed, 1 skipped, 2 failed.
- The other eight packages — 47 failed, 24 passed, 9 skipped.

All of those failures reproduce on unmodified `main`: I ran the same selection on both refs
and the sorted `FAILED` lists are identical (empty `diff`), with the same 47/24/9 totals.
They need QEMU binaries, serial ports, an Arduino CLI or NuttX images, none of which are
present here. `test_temp_disable_packages` fails for a separate reason — it expects
`ImportError` for the sibling packages, which the editable install from CONTRIBUTING makes
importable.

## Not included

Narrowing `except Exception`. `MessageQueue.put` already swallows a closed queue itself, and
the only escape I observed at teardown was `BrokenPipeError`, but I have not shown that list
to be exhaustive — so this PR stays one change. Happy to add it here if you would rather have
it together.

### DIFF
--- a/pytest-embedded/pytest_embedded/log.py
+++ b/pytest-embedded/pytest_embedded/log.py
@@ -154,7 +154,13 @@ class _PopenRedirectProcess(_ctx.Process):
 
     @staticmethod
     def _forward_io(msg_queue, logfile) -> None:
-        with open(logfile) as fr:
+        # Binary, because the file is still being written to: a text-mode read finalizes the
+        # decoder at every temporary EOF, and a UTF-8 sequence straddling that boundary raises
+        # UnicodeDecodeError. `except Exception` below would swallow it and this process would
+        # exit 0, cutting the output off for the rest of the session. Decoding here is not
+        # needed either - `MessageQueue.put` re-encodes with `to_bytes`, the pexpect buffer is
+        # bytes, and the serial transport already forwards `read_all()` undecoded.
+        with open(logfile, 'rb') as fr:
             while True:
                 try:
                     msg_queue.put(fr.read())  # msg_queue may be closed
--- a/pytest-embedded/tests/test_base.py
+++ b/pytest-embedded/tests/test_base.py
@@ -1010,3 +1010,84 @@ def test_multi_dut_no_data_loss(testdir):
 
     result = testdir.runpytest()
     result.assert_outcomes(passed=1)
+
+
+def _drain(q, deadline_s=10):
+    """Collect everything the queue yields until it stays quiet for a moment."""
+    import queue as _q
+    import time
+
+    chunks = []
+    deadline = time.monotonic() + deadline_s
+    while time.monotonic() < deadline:
+        try:
+            chunks.append(q.get(timeout=0.2))
+        except _q.Empty:
+            if chunks:
+                break
+    return b''.join(chunks)
+
+
+def test_forward_io_survives_split_multibyte_char(tmp_path):
+    """A multi-byte sequence split across two appends must not kill the forwarder.
+
+    The redirect log is written to while it is being read, so the reader keeps hitting a
+    temporary EOF. Decoding there turns any UTF-8 sequence straddling that boundary into
+    a UnicodeDecodeError, which the forwarder's `except Exception` swallows: the process
+    exits 0 and DUT output stops for the rest of the session.
+    """
+    import time
+
+    from pytest_embedded.log import MessageQueue, _PopenRedirectProcess
+
+    logfile = str(tmp_path / 'redirect.log')
+    open(logfile, 'wb').close()
+
+    q = MessageQueue()
+    p = _PopenRedirectProcess(q, logfile)
+    p.start()
+    try:
+        with open(logfile, 'ab') as fw:
+            fw.write(b'head\n')
+            fw.flush()
+        assert b'head\n' in _drain(q), 'nothing forwarded before the split'
+
+        # U+2014 EM DASH, torn in half: the reader reaches EOF between the two bytes.
+        with open(logfile, 'ab') as fw:
+            fw.write(b'dash \xe2')
+            fw.flush()
+        time.sleep(0.5)
+        assert p.is_alive(), f'forwarder died on a split sequence (exitcode={p.exitcode})'
+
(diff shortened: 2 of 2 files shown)


hfudev:
LGTM. Thank you for the fix. I'll do the release next Monday

PR #427, opened 2026-08-11
feat: add pytest-embedded-espemu service for esp-emu (CII-245)

shubhamdp:
### Description

Add a new `espemu` service that runs tests on [esp-emulator](https://github.com/espressif/esp-emulator), Espressif's emulator for ESP RISC-V series SoCs, instead of a real target — analogous to the existing `qemu` service:

```
pytest --embedded-services idf,espemu --target esp32c3
```

#### How it works:

- `EspEmuApp` merges the app's flash files into a single image (`esptool merge-bin`, no flash-size padding needed)
- `EspEmu` launches `esp-emu` with UART0 on stdio, so DUT I/O rides the standard DuplicateStdoutPopen` redirection — no serial ports or TCP bridges involved
- `hard_reset` raises `NotImplementedError`, which `IdfUnityDutMixin` already handles by re-triggering the Unity test menu
- New CLI options: `--espemu-image-path`, `--espemu-prog-path`, `--espemu-cli-args`, `--espemu-extra-args` (e.g. `"--net user,restrict=yes"` to isolate tests from LAN multicast noise)

- Supported targets: `esp32c3`, `esp32c6`, `esp32h2`, `esp32p4`, `esp32s31`; other targets fail with a clear error.

- Motivation: an emulator-backed DUT lets test suites that today run on QEMU or real hardware run anywhere with a single static binary — notably faster and with less CI setup.

- Naming: went with `espemu` to keep the service key, fixture, module, and CLI prefix identical (like `qemu`)

### Related
- https://github.com/espressif/esp-emulator

### Testing
- `pytest-embedded-espemu/tests/test_espemu.py` — pytester-based tests mirroring the qemu service's, against the `hello_world_esp32c3` fixture app (skipped when `esp-emu` is not on PATH): basic expect, multi-count with two emulator instances, and the unsupported-target error
- `pytest-embedded/tests/test_base.py` — service-to-classes parametrization extended with `idf,espemu` (covers the plugin wiring without needing the emulator binary)
- New `test-espemu` job in `test-python.yml` installs esp-emu via its upstream install script and runs the tests with coverage
- Full suite also verified in a clean venv (`bash foreach.sh install`) on macOS

- Also verified with the esp-matter project where this [unit-test-app](https://github.com/espressif/esp-matter/tree/main/examples/unit_test_app) is ran with this pytest-plugin using esp-emu

---

## Checklist

- [x] All CI checks (GH Actions) pass.

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -86,7 +86,7 @@ jobs:
           bash foreach.sh install
       - name: Download and install esp-emu
         run: |
-          curl -fsSL https://raw.githubusercontent.com/espressif/esp-emulator/main/install.sh | sh -s -- --version 0.38.0
+          curl -fsSL https://raw.githubusercontent.com/espressif/esp-emulator/main/install.sh | sh
           echo "$HOME/.local/bin" >> $GITHUB_PATH
 
       - name: Save PR number
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -54,6 +54,7 @@ jobs:
             --cov=pytest_embedded_idf \
             --cov=pytest_embedded_jtag \
             --cov=pytest_embedded_qemu \
+            --cov=pytest_embedded_espemu \
             --cov=pytest_embedded_serial \
             --cov=pytest_embedded_serial_esp \
             --cov=pytest_embedded_wokwi \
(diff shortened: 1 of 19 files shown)


.github/workflows/test-python.yml
```diff
@@ -68,6 +69,48 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          pip install -U pip
+          export PIP_EXTRA_INDEX_URL="https://dl.espressif.com/pypi"
+          pip install cryptography --prefer-binary
+          pip install -r requirements.txt
+          bash foreach.sh install
+      - name: Download and install esp-emu
+        run: |
+          curl -fsSL https://raw.githubusercontent.com/espressif/esp-emulator/main/install.sh | sh -s -- --version 0.38.0
```
mahavirj:
```suggestion
   curl -fsSL https://raw.githubusercontent.com/espressif/esp-emulator/main/install.sh | sh
```

Better to use latest version

shubhamdp: applied it locally

hfudev:
Thank you. All looks good to me. I'll merge it and release the 2.9.0 version next Monday

shubhamdp:
> Thank you. All looks good to me. I'll merge it and release the 2.9.0 version next Monday

Thank you!

PR #431, opened 2026-09-01
espemu: eFuses, flash encryption, esp32c5, and a working hard reset (CII-251)

Harshal5:
## Description

The qemu service can give a test an eFuse image, burn eFuses into it and boot an encrypted image. The esp-emu service could do none of that, and it had no `dut.serial` at all, so ESP-IDF tests that operate on device state had nothing to run against.

### eFuses

`--espemu-efuse-path` gives the emulator an eFuse image, creating a blank QEMU-compatible one when the file does not exist.

`execute_efuse_command()` runs espefuse against a second emulator instance started in download mode with its UART on a socket. A second instance is needed because the first one is booted into the firmware, and pyserial's `socket://` handler ignores modem control lines, so esptool cannot reset it into download mode.

### Flash encryption

`--encrypt` and `--keyfile` encrypt the merged image the way the qemu service does, with one difference: the command is `espsecure encrypt-flash-data --aes-xts`. Every target the emulator supports encrypts flash with XTS-AES

### esp32c5

esp-emu has accepted `--chip esp32c5` for a while, but the service rejected the target before launching it, so every esp32c5 app was unrunnable under `--embedded-services idf,espemu`. Nothing else in the service is target specific — the flash image comes from the app's own flash arguments — so the target list was the only gap.

### dut.serial.hard_reset()

ESP-IDF tests reach the chip through two channels: the console stream (`dut.expect`) and a control channel (`dut.serial`) that resets it, erases flash and burns eFuses. esp-emu now serves a control channel, so `hard_reset()` sends `reset` over it. The emulator process keeps running, so the dut's output stream, its expect history and its log all continue across the reset; relaunching the process would break all three.

The service checks `esp-emu --help` for `--control-tcp` before passing the flag, because an older binary exits on an unknown one. The operations that still need download mode raise `NotImplementedError` naming the operation that was wanted.

## Validation

Run against ESP-IDF master with `--embedded-services idf,espemu`:

- An encrypted image boots on an emulated target with the eFuses burned.
- esp32c5 collects and passes the same `esp_system` and `esp_hw_support` cases as the other targets.
- `esp_driver_uart`'s `test_uart_single_dev`, which used to fail at setup on `dut.serial`, now resets the chip six times and runs through to its test content.

## Dependency

`hard_reset()` needs an esp-emu build with `--control-tcp`, which no release carries yet. Without it the service behaves as it does today and `dut.serial.hard_reset()` raises with an explanation, so this can merge ahead of the emulator change.

---

## Checklist


pytest-embedded-espemu/pytest_embedded_espemu/espemu.py
```diff
@@ -55,22 +72,171 @@ def __init__(
...
+
+        child = subprocess.Popen(
+            [
+                self.espemu_prog_path,
+                '--chip',
+                self.target,
+                '--firmware',
+                self.image_path,
+                '--efuse',
+                self.efuse_path,
```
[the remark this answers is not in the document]
Harshal5: Fixed, amend into the `dut.serial` commit, the reload needs `control_command`, which that commit introduces.

The stale-machine half is real: the running emulator kept the eFuses it read at start, so a later reset missed the burn. It now reloads the image over the control channel with `efuse-load` once the burning instance exits. That resets the machine, which a burn needs on hardware anyway. Builds without `--control-tcp` are unaffected.

The overwrite-on-exit half doesn't reproduce: under `pytest-embedded` the image's `mtime` never changes after a run, so the emulator under test never writes its copy back. It is real in isolation, with a bare SIGTERM.

pytest-embedded/pytest_embedded/dut_factory.py
```diff
@@ -351,17 +354,21 @@ def _fixture_classes_and_options_fn(
...
                     EspEmu,
                 )
 
                 classes[fixture] = EspEmu
                 kwargs[fixture] = {
                     'msg_queue': msg_queue,
                     'espemu_image_path': espemu_image_path
-                    or os.path.join(app_path or '', build_dir or 'build', DEFAULT_IMAGE_FN),
+                    or os.path.join(
+                        app_path or '', build_dir or 'build', ENCRYPTED_IMAGE_FN if encrypt else DEFAULT_IMAGE_FN
```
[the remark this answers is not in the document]
Harshal5: Not a bug. The encrypt fixture is wrapped in `@multi_dut_argument`, which runs every value through `_str_bool` (plugin.py:366), so `false/no/n` arrive as `False`. Same reason the qemu branch above uses the expression unchanged.

pytest-embedded-espemu/pytest_embedded_espemu/serial.py
```diff
@@ -0,0 +1,76 @@
...
+        self.espemu.control_command(f'write-region {offset:#x} {file_path}')
+
+    def close(self) -> None:
+        """Nothing to close: the control channel is opened per command."""
+
+    def __getattr__(self, name: str) -> t.Any:
+        raise NotImplementedError(
+            f'esp-emu cannot do dut.serial.{name}() yet. Only hard_reset is '
+            f'implemented; flash and eFuse operations need the emulator in '
+            f'download mode with its UART on a socket.'
```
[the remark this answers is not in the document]
Harshal5: Fixed, the message was accurate when written, and went stale when `erase_flash`, `erase_region`, `erase_partition` and `write_flash_no_enc` landed. It now names what the requested operation needs instead of listing what the class has.

Harshal5:
@hfudev PTAL!

pytest-embedded-espemu/pytest_embedded_espemu/dut.py
```diff
@@ -27,6 +28,11 @@ def __init__(
         # target is known here as well; keep it optional for other app classes.
         self.target: str | None = getattr(self.app, 'target', None)
 
+        # `IdfSerial` is what ESP-IDF tests call for device state — a reset,
+        # an erase, an eFuse burn. Give them something that answers, so a test
+        # that needs one reports which operation it needed.
+        self.serial = EspEmuSerial(espemu, getattr(self, 'app', None))
```
hfudev: Hi, `serial` is also a fixture. Please refer to https://github.com/espressif/pytest-embedded/blob/0c1aa940ff9369b0004fa1d0ff2b206426d55bcf/pytest-embedded/pytest_embedded/dut_factory.py#L236

Harshal5: `EspEmuSerial` needs the `EspEmu` object to reach the control channel, but `serial_gn(_fixture_classes_and_options, msg_queue, app)` has no `espemu` parameter, every existing serial class is built from CLI options plus app alone.

Do you think we should add `espemu` to `serial_gn` and its fixture so it can be registered properly?

hfudev: the kwargs are passed through `_fixture_classes_and_options_fn`

later in the `serial_gn` the kwargs will be taken out

`    kwargs = _fixture_classes_and_options.kwargs['serial']`

Harshal5: PTAL again!

Harshal5: Change in https://github.com/espressif/pytest-embedded/pull/431/changes/e1e3f47f19aacfa2ec67897ce14aac0d996063c2

